PR #13, opened 2015-07-28
Don't promote other.xml usage — it is deprecated and leads to issues with ICS (settings-repository plugin)

develar:
@chashnikov Please review my changes. This decision was discussed with you some time ago. Due to lack of documentation our new CertificateManager starting to use other.xml. Now I have fixed CertificateManager and updated docs to avoid such situations.


chashnikov:
Looks ok but I don't have write access to this repository.

hhariri:
pinging @yole for access.

yole:
@chashnikov access granted

PR #16, opened 2015-09-16
change FindUsages getWordScanner not use a static instance.

vsch:
If it is static then there are multi-threading issues. When indexing multiple threads are running and overwriting the lexer state that the word scanner uses.

I copied the SimplePlugin find usages class and spent hours chasing index out range bugs. I also noticed that the SimplePlugin which I have installed in my plugin dev environment would once in a while cause an exception but I did not pay it much attention. The static word scanner is the source.


yole:
Thanks for the fix!

PR #17, opened 2015-12-14
Fix some links from .html to .md files and indent

vsch:
minor editing of stuff I noticed as I was poking around.


breandan:
Thanks, Vlad!

vsch:
My pleasure. I finally had some time to try the plugin on the project and found some bugs that I fixed. I also added translation from upsourse:// to https:// to the head branch for files. It makes it easier to follow the docs right out of the project. I'll spend more time going through the sdk docs after the next release.

PR #18, opened 2015-12-18
section which describes how to debug external build process updated accordingly to changes in IDEA 16 EAP

chashnikov:


breandan:
Thanks @chashnikov!

PR #21, opened 2016-02-28
fix include wrong file and icon dead link issue

zjhmale:
* should include `SimpleLanguage.java` instead of `Simple.bnf` in the `Define the language` section.
* fix dead link for the icon.

@YannCebron

### DIFF
--- a/tutorials/custom_language_support/language_and_filetype.md
+++ b/tutorials/custom_language_support/language_and_filetype.md
@@ -6,13 +6,13 @@ title: 2. Language and File Type
 ### 2.1. Define a language
 
 ```java
-{% include /code_samples/simple_language_plugin/src/com/simpleplugin/Simple.bnf %}
+{% include /code_samples/simple_language_plugin/src/com/simpleplugin/SimpleLanguage.java %}
 ```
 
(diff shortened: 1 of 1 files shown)


yole:
Why do you think the include file link is wrong? It's correct.

zjhmale:
@yole cause `SimpleFIleType.java` `SimpleTokenType.java` and `SimpleElementType.java` all dependent on `SimpleLanguage.java` and `Simple.bnf` just appearances in the `Grammar and Parser` chapter.

`SimpleLanguage.java` is the foundation element of this tutorial, and the section title is `Define the language` but not `Define the Parser and PSI for the language`.

zjhmale:
poke @yole

breandan:
@zjhmale You're right. [This mistake](https://github.com/JetBrains/intellij-sdk-docs/commit/b940673f47a3f102324f79e53cadc4fc111d083c#diff-a69807dcce170a2adf573d85726ef106) was introduced when we migrated from embedded code snippets to Liquid tags. Thanks for the PR!

zjhmale:
you're welcome :)

PR #29, opened 2016-12-27
Fix JDK version so that it is in sync with the screenshot above

bperel:


kingburrito666:
They say real heroes don't exist...

ignatov:
Thanks

PR #37, opened 2017-05-25
corrected broken links in FAQ section

chashnikov:
Links to old devnet forums returns 404 error now, so they must be replaced by links to the corresponding topics in the new forum.


chashnikov:
BTW the first link in FAQ points to the old devnet forum as well (https://devnet.jetbrains.com/community/idea/open_api_and_plugin_development). It currently properly redirects to the new forum, but maybe we should use the direct link instead?

ignatov:
I suppose it's better to use new links instead of redirections.

citizenmatt:
Thanks. I've updated the first link in the FAQ, too.

PR #45, opened 2017-08-19
Update snippet code

fmodesto:
This makes reference to a future code, requires `com.simpleplugin.psi.impl.SimplePsiImplUtil` which is not implemented until step 6. Also the explanation seems to make reference to the snippet show in the screenshot bellow.


ignatov:
Thanks

PR #50, opened 2017-09-21
Removes vendor logo

abeatte:
Vendor logo is no longer used and including it in the xml produces a lint warning.


citizenmatt:
Brilliant, thanks!

PR #51, opened 2017-10-01
Dependencies to specific plugins

markiewb:
Based on my observation in https://github.com/markiewb/idea-webstorm-pluginsetup-example and https://intellij-support.jetbrains.com/hc/en-us/community/posts/206762465-How-do-I-make-use-of-the-javascript-psi-


citizenmatt:
Thanks!

PR #52, opened 2017-10-01
Add link to OSS license page

markiewb:
Some OSS plugins may be eligible to get free licenses to check the plugin compatibility against different products


citizenmatt:
Nice one, thanks!

PR #54, opened 2017-10-20
Update language_and_filetype.md

sebastienvermeille:
That way we know where to place this xml string... I added it in <extensions-points> and of course it doesn't work :/


ignatov:
Thanks

sebastienvermeille:
@ignatov you are welcome thank to you for providing such documentation ! 👍

PR #56, opened 2017-10-21
Remove orphaned closing tag from plugins.xml excerpt

calexHG:
There was an orphaned `action` closing tag. This repurposes it as a closing tag for the `actions` tag. Fixes the previous fix.


ignatov:
Thanks.

PR #68, opened 2017-12-05
Close vendor tag in sample plugin.xml

jonnybot0:
Minor bug, but the sample code is broken.


citizenmatt:
Oops! Good spot, thanks for the fix!

PR #72, opened 2017-12-25
IntelliJ Platform 2018.1 api change - SVNKit is removed

pchel-:


chashnikov:
You're supposed to use one of the supported patterns in the first column, see the comment at the beginning of the file for details.

pchel-:
OK, thanks. See #73.

PR #85, opened 2018-02-13
Remove passing a file extension explicitly.

dmitvitalii:
Hello!

In this code, the `SimpleFileType` instance passed as the first argument and the `"simple"` file extension as the second parameter.

As far as I can see in FileTypeManagerImpl:263:
```java
@Override
public void consume(@NotNull FileType fileType) {
  register(fileType, parse(fileType.getDefaultExtension()));
}
```
There is no need to declare it explicitly as it is taken by calling `SimpleFileType.#getDefaultExtension()` by default, so this parameter is redundant.

To make tutorial more simple, I suggest this removal: seeing the second parameter is a bit confusing when we're passing the same value, returned from `getDefaultExtension()`.
Since the tutorial shows the basics, not a flexibility of an API, the parameter can be removed.

Thanks!


yole:
Thanks!

PR #98, opened 2018-05-22
fixed external github code sample link

4rg0n:
Fixed link for PsiNavigationDemoAction.java


yole:
The psi_demo sample has since been moved to the correct directory, so this change is no longer needed.

PR #100, opened 2018-06-01
Quotes in java string

MrTsepa:
I think quotes should look like this for this file to be compilable.


yole:
This is a testdata file, it does not need to be compilable. The quotes are correct.

PR #102, opened 2018-06-27
IJSDK-50: remove checkout_and_build_community; point to intellij-community/README.md instead

JohnHake:
I removed the intellij-sdk-docs page about building IntelliJ IDEA. All previous links to that page (including _SUMMARY.MD, the menu,) now link to https://github.com/JetBrains/intellij-community/blob/master/README.md.


_SUMMARY.md
```diff
@@ -19,7 +19,7 @@
         * [Publishing Your Plugin](tutorials/build_system/deployment.md)
     * [Using DevKit](basics/getting_started/using_dev_kit.md)
         * [Setting Up a Development Environment](basics/getting_started/setting_up_environment.md)
-        * [Check Out And Build Community Edition](basics/checkout_and_build_community.md)
+        * [Check Out And Build Community Edition](https://github.com/JetBrains/intellij-community/blob/master/README.md)
```
yole: I think we should remove this reference entirely, because it's not part of this documentation and not a required step for setting up a plugin development environment.

basics/getting_started.md
```diff
@@ -14,7 +14,7 @@ In this section:
     * [Publishing Your Plugin](/tutorials/build_system/deployment.md)
 * [Developing plugins using DevKit](getting_started/using_dev_kit.md)
     * [Setting Up a Development Environment](getting_started/setting_up_environment.md)
-    * [Check Out And Build Community Edition](basics/checkout_and_build_community.md)
+    * [Check Out And Build Community Edition](https://github.com/JetBrains/intellij-community/blob/master/README.md)
```
yole: This can be removed too.

JohnHake:
I would prefer to leave at least some of these references in the SDK documentation. 
* Regardless of whether building IntelliJ CE is a required step, the instructions are still very valuable to SDK consumers. If having a link in _SUMMARY.md is too much visibility with not enough context, I'd prefer to remove that link but leave the link in [setting_up_environment.md](https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/setting_up_environment.html). (Which could stand to be clearer about whether buildiing IntelliJ CE is required, and if not, explain the alternative procedure.)
* intellij-community/README.md doesn't get much search engine visibility. While this is a separate problem, at least in the interim I would like to make it easier to find for SDK consumers.
* It was my intention to also link to intellij-community/README.md from [JetBrains.org](https://www.jetbrains.org/pages/viewpage.action?pageId=983225) so the Contribute and SDK sites both point to one location for building IntelliJ CE from source. Given your comments, should I rethink this plan for JetBrains.org?

yole:
Sure, please do leave the reference in setting_up_environment.md. I just don't think it's appropriate to have it in the TOC.

Linking to the README from jetbrains.org is perfectly fine too.

JohnHake:
How would the source paths for **IDEA jdk** be defined if one chooses _not_ to get IntelliJ IDEA CE source code? setting_up_environment.md is still vague on this point.

yole:
If you use devkit and choose not to checkout the CE source code, then basically you won't have sources. We used to have a [page](http://www.jetbrains.org/display/IJOS/Download) with links to source archives, but it hasn't been updated in ages. (The source archives are still being published, but there is no way for the user to find the download link.)

JohnHake:
Ok, then I think the latest commit covers the points we discussed.

PR #104, opened 2018-08-01
Add org.apache.batik.script.InterpreterFactory#createInterpreter to the API change

forantar:
The API changed since upgrading to batik 1.10


chashnikov:
Thank you!
And what about 'Access to unresolved class org.apache.batik.script.rhino.RhinoInterpreter' compatibility issue?

forantar:
I've stub'ed RhinoInterpreter, commit 4864180.

PR #107, opened 2018-08-08
Do not include provided kotlin runtime in plugin

Recognized:
Intellij platform now provides com.jetbrains:kotlin-reflect, com.jetbrains:kotlin-runtime, com.jetbrains:kotlinx-coroutines-core, com.jetbrains:kotlinx-coroutines-core jars.

### DIFF
--- a/tutorials/kotlin.md
+++ b/tutorials/kotlin.md
@@ -64,7 +64,17 @@ repositories {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


yole:
The current version of kotlinx.coroutines is 0.24. It's a prerelease version, and as such, we make no guarantees regarding its binary or source compatibility. If a plugin uses kotlinx.coroutines, it should bundle it.

PR #108, opened 2018-08-14
Refactor register_actions code and documentation.

JohnHake:
Testing of doc revisions and plugin refactoring is complete.


code_samples/register_actions/resources/META-INF/plugin.xml
```diff
@@ -63,21 +63,45 @@
...
+      <add-to-group group-id="EditorPopupMenu" anchor="first"/>
+      <action class="org.jetbrains.tutorials.actions.SimplePopDialogAction" id="org.jetbrains.tutorials.actions.CustomGroupedAction"
+              text="A Custom Grouped Action" description="Custom Grouped Action Demo"/>
     </group>
-    <group id="BaseActionGroup" class="org.jetbrains.tutorials.actions.BaseActionGroup" popup="true"
-           text="ActionGroup Demo" description="Extending AnAction Demo">
-      <add-to-group group-id="ToolsMenu" anchor="first"/>
+    <!-- DynamicActionGroup demonstrates declaring an action group without a static action declaration.
+     An action is added to the group programmatically in the DynamicActionGroup implementation. The group is added
+     at the top of the tools menu. -->
```
yole: at the bottom (you're using `anchor="last"`)

code_samples/register_actions/src/org/jetbrains/tutorials/actions/SimplePopDialogAction.java
```diff
@@ -0,0 +1,78 @@
...
+ * The only action this class performs is to provide the user with a popup dialog as feedback.
+ * Typically this class is instantiated by the IntelliJ Platform framework based on declarations
+ * in the plugin.xml file. But when added at runtime this class is instantiated by an action group.
+ * @see com.intellij.openapi.actionSystem.AnAction
+ * @see com.intellij.openapi.actionSystem.AnActionEvent
+ */
+public class SimplePopDialogAction extends AnAction {
+
+  // Capture the generic message icon for display in the dialog.
+  private static final Icon ourIcon = Messages.getInformationIcon();
```
yole: Having static fields in `AnAction`-derived classes is a great way to introduce memory leaks, so we should not show this in our sample code, even though this particular usage is harmless. You can simply inline this field.

code_samples/register_actions/src/org/jetbrains/tutorials/actions/SimplePopDialogAction.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Gives the user feedback when the dynamic action menu is chosen.
+   * Pops a simple message dialog.
+   * @param anActionEvent Event received when the associated menu item is chosen.
+   */
+  @Override
+  public void actionPerformed(@NotNull AnActionEvent anActionEvent) {
+    // Using the event, create and show a dialog
```
yole: I actually like Anna's original example more in this place. It demonstrated accessing data from `AnActionEvent`, which is a fundamental part of using actions, and your example shows only working with the action presentation, and does that in a way which is unlikely to be useful for plugin developers.

code_samples/register_actions/resources/META-INF/plugin.xml
JohnHake: Careless of me. I changed the menu locations but not the comment.

code_samples/register_actions/src/org/jetbrains/tutorials/actions/SimplePopDialogAction.java
JohnHake: I will change the field.
I'd like to better understand why static fields might introduce memory leaks. Perhaps we can discuss off-line.

code_samples/register_actions/src/org/jetbrains/tutorials/actions/SimplePopDialogAction.java
JohnHake: I like Anna's `anAction` implementation as well. Here's my thinking...
IMO a tutorial like `register_actions` should focus on the ways actions/groups get registered and how they impose context requirements. (`psi_demo` already does a nice job of demonstrating data access from `AnActionEvent`.)
`SimplePopDialogAction` demonstrates determining context, and with `CustomDefaultActionGroup` shows how to layer more restrictive contexts. Maybe there are more relevant contexts; I'm open to suggestions. That said, there's virtue in `SimplePopDialogAction` having modest context requirements because it can be used/reused throughout the code sample. Previously most of the example actions in `register_actions` did not give  users feedback when a menu was chosen. I think providing user feedback (however trivial) is important in this tutorial, and `register_actions` now uses `SimplePopDialogAction` to do that in all cases.

yole: Providing user feedback is definitely better than having empty actions as in Anna's version of the tutorial. However, I do think that working with action data is an essential part of working with actions, and should be demonstrated in the core actions tutorial, not in a more topic-specific tutorial such as `psi_demo`. You can easily combine the two and show some data taken from the actionevent in the dialog.

JohnHake: I have committed what I hope is a good compromise - using the `Navigatable` object opportunistically in `SimplePopDialogAction.actionPerformed()`. I have updated the  _Registering an Action_ documentation to discourage using class fields in `AnAction` subclasses.

PR #109, opened 2018-08-22
IJSDK-431 Rough draft of new doc page for artifacts repository.

JohnHake:
(cherry picked from commit 78069e92b5febaf483af9f1c1e868c913551cb68)

Please review this rough draft of reference_guide/intellij_artifacts.md in particular. The document contains several questions which are marked with "@nik". Feel free to edit / add comments as needed; all comments and suggestions are welcome.

/cc @chashnikov
/cc @shalupov

### DIFF
--- a/reference_guide/intellij_artifacts.md
+++ b/reference_guide/intellij_artifacts.md
@@ -1,43 +1,35 @@
 ---
-title: IntelliJ IDEA Community Edition Artifacts Repository
+title: IntelliJ Platform Artifacts Repository
 ---
 
-@nik - a few questions in the document. The one systemic improvement I'd like to make is about nomenclature. Why is `artifactId` used
-in the Artifact Description, but `name` is used as a key in the example Dependency statement? I feel like I need to provide a set
-of bullet items after each example to translate the terms.
+JetBrains maintains a public repository that hosts artifacts related to the IntelliJ Platform, such as binaries and source code. This 
+repository makes artifacts more accessible for plugin authors.
 
-JetBrains maintains a public repository for hosting IntelliJ IDEA Community Edition artifacts such as binaries and source code. This 
-repository makes artifacts more accessible for plugin writers.
-
-@nik - Does this repository replace a previous system for external plugin developers, or were the artifacts just not available?  
-@nik - Should we state on a public page that there will be additional products in the future - any concern about leading with our chin? 
-
-Multiple types of content are hosted:
+The IntelliJ Platform Artifacts Repository contains:
 * [Releases repository](https://www.jetbrains.com/intellij-repository/releases/) for release and EAP versions.
 * [Snapshots repository](https://www.jetbrains.com/intellij-repository/snapshots/) for nightly successful 
   builds of each branch.
-* [Dependencies](https://jetbrains.bintray.com/intellij-third-party-dependencies) of individual modules from the IntelliJ IDEA project.  
+  
+More artifacts will be hosted in the future.
 
-### How to Use the Repository
-There are three steps to use the repository:
-1. Add the corresponding repository URL to a maven or gradle script.
-2. Specify the description for the desired artifact.
-3. Specify the build type and version for the artiface.
- 
+[Dependencies](https://jetbrains.bintray.com/intellij-third-party-dependencies) of individual modules from the IntelliJ
+Platform are hosted at the bintray repository. These artifacts should not be used directly, but a link to this 
+repository should be added to pom.xml/build.gradle files if individual modules from the IntelliJ Platform Artifacts Repository are used.    
+
+### How to Use the IntelliJ Platform Artifacts Repository
+There are two steps to using the repository:
+1. Add the corresponding repository URL to a Maven or Gradle script.
+2. Specify the Maven coordinates for the artifact. 
  
 #### Add the Repository URL 
-The corresponding URL for the desired artifact needs to be added to a maven or gradle script:
+The corresponding URL for the desired artifact needs to be added to a Maven or Gradle script:
 * For release or EAP versions, use https://www.jetbrains.com/intellij-repository/releases 
 * For snapshot or EAP snapshots, use https://www.jetbrains.com/intellij-repository/snapshots
-* For dependencies of individual modules from the IntelliJ IDEA Community Edition Project, use https://jetbrains.bintray.com/intellij-third-party-dependencies 
+* For dependencies of individual modules from the IntelliJ Platform, use https://jetbrains.bintray.com/intellij-third-party-dependencies 
 
-@nik - the dependencies URL leads
-to a page with `org/` which leads to `jetbrains/` to `intellij/` to `deps/` which seems like a lot to wade through. Should
-the URL include these subdirectories?
-
-#### Specify the Artifact Description
-Describing the details of a desired artifact is done with parameters to specify the IDs, type, and packaging of the artifacts:
-* Cross-platform zip distributions of IntelliJ IDEA Community Edition:
+#### Specify the Artifact
+Describing the desired artifact is done with Maven coordinates:
+* Cross-platform zip distributions of IntelliJ Platform artifacts:
   * groupId = com.jetbrains.intellij.idea
   * artifactId = ideaIC
   * classifier = ""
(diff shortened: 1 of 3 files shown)


reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+Where:
+* `com.jetbrains.intellij.platform` corresponds to the `groupId` for modules in the Artifact Description section.
+* `jps-model-serialization` corresponds to the `artifactId` in the Artifact Description section.
+* `182.2949.4` specifies the artifact version - see the Artifact Version section.
+* `jps-model-impl` corresponds to the module `artifactId` in the Artifact Description section.
+* **Note:** the artifact version must match in both statements.
+
+@nik - Are the statements related in terms of both specifying `jps-model-`? Is there an example where only one module
```
chashnikov: Yes, these two artifacts are related, `jps-model-serialization` declares some API and `jps-model-impl` provides its implementation. Currently if you want to use classes from `jps-model-serialization` you also need to manually add dependency on `jps-model-impl`. Probably we'll get rid of this later and dependency will be added automatically.

As an example with only one artifact you can use `util` artifact from the same `com.jetbrains.intellij.platform` group.

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,131 @@
...
+* `name` corresponds to the `artifactId` in the Artifact Description section.
+* `version` specifies the artifact version - see the Artifact Version section.
+* `ext` corresponds to the `packaging` in the Artifact Description section.
+
+@nik - Is `idea_dep` being declared as this particular artifact dependency?
+
+Here is an example of a [build.gradle](https://github.com/shalupov/idea-cloudformation/blob/9007023afa187a1fb8b45c3ca66d5a51f86b795c/build.gradle)
+that is currently in use.
+
+#### Gradle Example for an Iindividual Module from the IntelliJ IDEA project
```
zolotov: Typo in individual

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,107 @@
...
+  * groupId = com.jetbrains.intellij.platform
+  * artifactId = jps-model-serialization
+  * classifier = ""
+  * packaging = jar
+
+For each artifact there are several versions available. The version can be specified in one of several ways:
+* A branch build is specified as _BRANCH.BUILD[.FIX]_. For example, a branch build such as `141.233`, or a branch build with a fix such as `139.555.1`
+* Release numbers are specified as _MAJOR[.MINOR][.FIX]_. For example `14`, or `14.1`, or `14.1.1`
+* A snapshot of a branch is specified as _XXX-SNAPSHOT_. For example `142-SNAPSHOT`
+* A snapshot of an early release is specified as _BRANCH-EAP-SNAPSHOT_. For example `141-EAP-SNAPSHOT`
```
chashnikov: "an early release" looks obscure. I would write something like "a snapshot of a branch from which the next EAP/release build will be produced".

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,107 @@
...
+* Individual modules from the IntelliJ Platform with their dependencies. For example, to specify the jps-model-serialization module:
+  * groupId = com.jetbrains.intellij.platform
+  * artifactId = jps-model-serialization
+  * classifier = ""
+  * packaging = jar
+
+For each artifact there are several versions available. The version can be specified in one of several ways:
+* A branch build is specified as _BRANCH.BUILD[.FIX]_. For example, a branch build such as `141.233`, or a branch build with a fix such as `139.555.1`
+* Release numbers are specified as _MAJOR[.MINOR][.FIX]_. For example `14`, or `14.1`, or `14.1.1`
+* A snapshot of a branch is specified as _XXX-SNAPSHOT_. For example `142-SNAPSHOT`
```
chashnikov: "A snapshot of a branch" looks too vague. I would write something like "The latest nightly build from a branch" instead.

chashnikov:
Thank you, this looks good!

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,107 @@
...
+  * artifactId = ideaIC
+  * classifier = sources
+  * packaging = jar
+* Individual modules from the IntelliJ Platform with their dependencies. For example, to specify the jps-model-serialization module:
+  * groupId = com.jetbrains.intellij.platform
+  * artifactId = jps-model-serialization
+  * classifier = ""
+  * packaging = jar
+
+For each artifact there are several versions available. The version can be specified in one of several ways:
```
shalupov: Also versions are available at https://www.jetbrains.com/intellij-repository/releases and https://www.jetbrains.com/intellij-repository/snapshots, it's not immediately obvious for everyone

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,107 @@
...
+More artifacts will be hosted in the future.
+
+[Dependencies](https://jetbrains.bintray.com/intellij-third-party-dependencies) of individual modules from the IntelliJ
+Platform are hosted at the bintray repository. These artifacts should not be used directly, but a link to this 
+repository should be added to pom.xml/build.gradle files if individual modules from the IntelliJ Platform Artifacts Repository are used.    
+
+### How to Use the IntelliJ Platform Artifacts Repository
+There are two steps to using the repository:
+1. Add the corresponding repository URL to a Maven or Gradle script.
+2. Specify the Maven coordinates for the artifact.
```
shalupov: Could we specify here and everywhere else that the main method of consuming those repositories is https://github.com/JetBrains/gradle-intellij-plugin and not directly? cc @chashnikov

JohnHake: @shalupov  - Thanks for your review.
Yes, I can clarify. I'll add a lead-in sentence to the effect that the two steps describe using the repositories by adding information to a project's Gradle build script, which is supported by gradle-intellij-plugin.

reference_guide/intellij_artifacts.md
JohnHake: @shalupov - Thanks for the feedback. I'll add clarification to the lead-in sentence..."For each artifact _at the Repository URLs_ there are several...". The added clarification will link to the preceding "Add the Repository URL" section.

JohnHake:
If this content looks ok, would someone please merge it?

reference_guide/intellij_artifacts.md
chashnikov: It loos like this wasn't fully resolved. Currently we have
> Artifacts in the repositories are utilized by adding information to a project’s build.gradle file. See the gradle-intellij-plugin for more information about Gradle support.

This may be misleading. You don't need to use gradle-intellij-plugin to access artifacts for individual modules, it won't help you with that. Actually we have two kinds of artifacts in there repositories. The first one is full zip distributions of IntelliJ IDEA (and theirs sources). These artifacts aren't supposed to be accessed directly, they are used inside gradle-intellij-plugin (see [my comment](https://github.com/JetBrains/intellij-sdk-docs/pull/109#discussion_r214303721) in this pull request). The second one is individual modules from IntelliJ Platform. They may be used directly and they aren't quite related to gradle-intellij-plugin.

Maybe we should simply remove description of the first kind of artifacts from the public page? (The first two items in 'Specify the Artifact' section, and 'Gradle Example for IntelliJ IDEA Community Edition Distribution' section.) And replace them by a small note like "These repositories also provides cross-platform full distributions of IntelliJ IDEA which are used by gradle-intellij-plugin and aren't supposed to be accessed directly"? This will greatly simplify the document.

@JohnHake @zolotov @shalupov what do you think?

reference_guide/intellij_artifacts.md
```diff
@@ -0,0 +1,111 @@
...
+
+There are two types of information needed to use a repository:
+1. Specify the corresponding repository URL for the artifact.
+2. Specify the Maven coordinates for the artifact. 
+ 
+### Add the Repository URL 
+The corresponding URL for the desired artifact needs to be added to a Maven or Gradle script:
+* For release or EAP versions, use https://www.jetbrains.com/intellij-repository/releases 
+* For snapshot or EAP snapshots, use https://www.jetbrains.com/intellij-repository/snapshots
+* For dependencies of individual modules from the IntelliJ Platform, use https://jetbrains.bintray.com/intellij-third-party-dependencies
```
chashnikov: I would replace 'use' by 'also use' in the third item. As stated above, `intellij-third-party-dependencies` is an additional repository, and if you want to use individual modules you still take them from `releases` or `snapshot` repository.

reference_guide/intellij_artifacts.md
zolotov: +1 for removing the first kind of artifacts from the public page and for adding `These repositories also provides cross-platform full distributions of IntelliJ IDEA which are used by gradle-intellij-plugin`.

Not sure about `aren't supposed to be accessed directly`, as they still can be downloaded directly if needed modules are not published separately yet.

chashnikov: @JohnHake so what do you think about this?

PR #111, opened 2018-08-23
ijsdk-85 - Revised setting up php environment in IntelliJ. Refreshed all images.

JohnHake:
Please review.

New section w/image describing how to add php libraries to class path of project sdk.

Check section _1.2 Add Php Library to the SDK Classpath_ - even after reading this bug I'm not clear if one or both php libs should be added to the classpath.

Search for "@anet" to find specific questions.

/cc @akozlova


phpstorm/setting_up_environment.md
```diff
@@ -3,37 +3,57 @@ title: Setting-up the Environment for PhpStorm Plugin Development
...
-
+1. Use IntelliJ IDEA Ultimate as the IDE with the PHP plugin installed.
+2. Use PhpStorm as the IDE.
+
+The choice affects how you will configure an SDK for developing a plugin. Using IntelliJ IDEA Ultimate as the IDE requires basing the
+SDK for your project on your installed version of IntelliJ IDEA. Using PhpStorm as the IDE requires basing the SDK for
+your project on your installed version of PhpStorm. Choosing an installed IDE as the basis for a project is described on the [Setting Up a Development Environment](/basics/getting_started/setting_up_environment.md)
+page.
+
+@anet - is IDEA Ultimate required, or could a developer use IntelliJ CE?
```
yole: IntelliJ IDEA Community Edition does not support the PHP plugin. The IDE in which the plugin is developed can be IJ CE, but the IDE used for testing needs to be IJ Ultimate.

phpstorm/setting_up_environment.md
```diff
@@ -3,37 +3,57 @@ title: Setting-up the Environment for PhpStorm Plugin Development
...
+
+## How to use the PhpStorm OpenAPI Library in IntelliJ
+
+This section explains how to configure the IntelliJ Platform for using the PhpStorm OpenAPI.
```
yole: This is the regular process of adding a plugin dependency, covered here: http://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_dependencies.html

phpstorm/setting_up_environment.md
```diff
@@ -3,37 +3,57 @@ title: Setting-up the Environment for PhpStorm Plugin Development
...
+
+> Note: the OpenAPI library is available for PhpStorm 6 and above.
+
+### 1. Adding the Php Library
+First the Php library must be added to your plugin's project settings.
+
+#### 1.1 Add Php Libraries to a Module
```
yole: This step is not needed at all (nor is 1.3). It's enough to add the library to the SDK classpath.

cedricziel:
I feel like this document is a bit too complicated. - Setting up such a project with gradle is a breeze - why not push the gradle-plugin instead of this - sorry - setup-mess?

With gradle, it's very exact and no need to change the library path and fiddle with scopes.

Examples include the [Symfony Plugin](https://github.com/Haehnchen/idea-php-symfony2-plugin/blob/master/build.gradle) (single module) and the [TYPO3 CMS Plugin](https://github.com/cedricziel/idea-php-typo3-plugin) (multi-module). (Disclosure: I migrated both projects to gradle, because i've found setting the projects up to be a horrible experience)

I've always found it horrible for new contributors to set-up the plugin project. With gradle, all they have to to is import the gradle file, check auto-import and the project is ready to roll.

JohnHake:
Revised content per feedback.

phpstorm/setting_up_environment.md
```diff
@@ -2,42 +2,30 @@
...
+The choice affects how you will configure an SDK for developing a plugin. Using IntelliJ IDEA Ultimate as the IDE requires basing the
+SDK for your project on your installed version of IntelliJ IDEA. Using PhpStorm as the IDE requires basing the SDK for
+your project on your installed version of PhpStorm. Choosing an installed IDE as the basis for a project is described on 
+the [Setting Up a Development Environment](/basics/getting_started/setting_up_environment.md) page.
 
-The choice affects how you will configure SDK for the plugin. In the first case you need to specify current installation of IntelliJ IDEA as SDK and in the second case you need to specify current installation of PhpStorm. This step is described in [Getting Started with Plugin Development](/basics/getting_started.md).
+> IntelliJ IDEA Community Edition does not support the PHP plugin. The IDE for developing a PhpStorm plugin can be IntelliJ IDEA Community Edition, but the IDE used for testing must be IntelliJ IDEA Ultimate.
 
-## How to use OpenAPI library
+### How to use the PhpStorm OpenAPI Library for IntelliJ Plugin Devlopment
```
yole: I don't think we should mention "IntelliJ Plugin Development" here - the article covers plugins for PhpStorm, not for IntelliJ.

phpstorm/setting_up_environment.md
```diff
@@ -2,42 +2,30 @@
...
 
 > OpenAPI is available for PhpStorm 6 and above.
 
-This section explains how to configure the IntelliJ Platform for using PhpStorm OpenAPI.
+The [Plugin Dependencies](/basics/plugin_structure/plugin_dependencies.md) instructions cover adding dependencies to the
+DevKit as well as Gradle development environments. Follow the instructions to:
 
-### Adding library to a module
+* Add the `php-openapi.jar` and `php.jar` libraries to your plugin project. These libraries are located
```
yole: The information about .jar names and locations is only relevant for projects using the DevKit based enviroment. I'd reformulate this paragraph as "Add PHP OpenAPI classes to the classpath of your plugin". For Gradle-based setup, it's also worth mentioning that the plugin ID to add to build.gradle is `com.jetbrains.php`.

phpstorm/setting_up_environment.md
akozlova: >There are two ways to develop plugins for PhpStorm:
I'd rather point out that there are two possibilities to target IDEs. Otherwise one could read it as PhpStorm can be used to develop the plugin and that's not the case as it doesn't support java. The development IDE could be either Ultimate or Community

JohnHake: Thanks for the feedback Anna. I made changes to clarify the difference between the target IDE for a plugin, and the IDE used to develop the plugin.

phpstorm/phpstorm.md
```diff
@@ -2,10 +2,13 @@
 title: PhpStorm Plugin Development
 ---
 
-Plugins for PhpStorm are developed in Java using IntelliJ IDEA (Community Edition is sufficient in this case). You will also need a copy of PhpStorm to develop against. The [PsiViewer plugin](https://plugins.jetbrains.com/plugin/227-psiviewer) will also be useful.
+Plugins for PhpStorm are developed in Java using either IntelliJ IDEA or PhpStorm. IntelliJ IDEA Community Edition does not support the PHP plugin.
+IntelliJ IDEA Community Edition can be used to develop a PhpStorm plugin, but IntelliJ IDEA Ultimate must be used for testing.
+ 
+You will also need a copy of PhpStorm to develop against. The [PsiViewer plugin](https://plugins.jetbrains.com/plugin/227-psiviewer) will also be useful.
```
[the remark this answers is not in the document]
JohnHake: Let's clarify what we mean by _targeting a product_ vs. _which product can be used for development_ vs. _how to define the SDK classpaths_.  Here is my understanding:
* **Target:** These pages are about developing a plugin that targets PhpStorm. Discussion about a PHP plugin targeting (extending) IntelliJ IDEA seems to be confusing matters, so for now let's remove it.
*  **IDE:** PhpStorm plugins are written in Java, so IntelliJ IDEA is required. 
   * What are the PHP development limitations in IJ CE? For example, will IJ CE build a PhpStorm plugin project against an SDK that includes PHP libraries in the class path? 
   * When you say testing must be done in IntelliJ IDEA Ultimate rather than CE, does that mean CE does not support debugging a PhpStorm plugin?
   * If testing should be done in PhpStorm anyway, then let's explicitly say the product is also required.
*  **SDK:** If PhpStorm is required for testing, then why wouldn't a PhpStorm plugin project SDK always be based on the installed version of PhpStorm? (Rather than the installed version of IntelliJ.)

If you would please clarify these points then I will update phpstorm.md and setting_up_environment.md.  TIA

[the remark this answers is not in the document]
yole: We probably should come up with a consistent set of terms for the two IDEs that come into play when developing a plugin: the IDE in which you write the plugin code ("development IDE") and the IDE which you run to test the plugin ("target IDE").

As the development IDE, you can use IntelliJ IDEA Community Edition; there are no limitations.

As the target IDE, you can't use IntelliJ IDEA Community Edition, because it won't load the PHP plugin, and therefore your plugin which depends on PHP won't work either. You can use IntelliJ Ultimate with the PHP plugin, but it's better if you use PhpStorm, because in that way you avoid the risk of using some APIs which aren't available in PhpStorm.

If you're using the devkit project set up, you can create an IntelliJ Platform plugin SDK based on the installation of any target IDE, including PhpStorm, and indeed you should create it based on an installation of PhpStorm.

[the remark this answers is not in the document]
JohnHake: Here's my understanding of developing PHP and PhpStorm plugins...
Development IDE: the IDE in which you write and build the plugin code.
Test IDE:  the IDE in which you are able to install and run the plugin for testing.
Target IDE: the IDE in which a _user_ will run the plugin.

JetBrains Product | Development IDE | Test IDE | Target IDE
------------ | ------------- | ------------- | -------------
IntelliJ CE | Y | N (1) | N (1)
IntelliJ UE | Y (2) | Y (2) | Y (2,3)
PhpStorm | N (4) | Y  | Y (5)

1. IntelliJ CE doesn't support the PHP plugin
2. Requires PHP plugin. 
3. Best practice is to base the plugin project SDK on an IntelliJ installation.
4. PhpStorm does not support Java development.
5. Best practice is to base the plugin project SDK on PhpStorm installation.

How do developers of PhpStorm plugins typically debug their code? Is debugging done within a development instance of IntelliJ?

PR #114, opened 2018-09-05
Add information about feature extractor

LChernigovskaya:
According to https://youtrack.jetbrains.com/issue/MP-1829 the information about the feature extractor was added.


yole:
@JohnHake Could you please proofread the English text and make changes as necessary? Thanks!

JohnHake:
Nice job! Just a few minor changes here and there.
I noticed the module type image is not used. I did not add it to plugin_repository/feature_extractor.md
See JohnHake/intellij-sdk-docs:mp-1829 for changes.

LChernigovskaya:
Thanks a lot!

Now IDEs don't advertise plugins with module type feature, that's why the image which I have is not a good example. I think I will remove this image. When it is fixed, I will add a new image.

yole:
I've renamed the page to "Plugin Recommendations", rewritten the intro section to describe the actual scenarios when plugin recommendations come into play, and removed most of the references you added to other pages (because in almost all cases it's not something that plugin developers need to worry about - and most definitely not the first thing they would need to worry about.)

PR #116, opened 2018-09-07
ijsdk-461: Refactored prerequisites.md

JohnHake:
Replaced redundant content with links to existing content. Deleted redundant images.

/cc @yole


yole:
Would be nice to convert the tutorial to a Gradle project.

PR #119, opened 2018-09-17
Proper way to add plugin dependencies in gradle

thsaravana:
Adding this: `plugins 'org.jetbrains.kotlin:1.2.30'` to build script doesn't work and gives the following error message.
**Cannot find plugin org.jetbrains.kotlin:1.2.30 at https://plugins.jetbrains.com**

We need to give full version of the plugin like `plugins 'org.jetbrains.kotlin:1.2.70-release-IJ2018.2-1'` to work.

An easier way to do this is to simply mention `"Kotlin"`

### DIFF
--- a/basics/plugin_structure/plugin_dependencies.md
+++ b/basics/plugin_structure/plugin_dependencies.md
@@ -21,7 +21,7 @@ dependencies (e.g. Kotlin), you need to perform the following steps:
 
 ```groovy
 intellij {
-    plugins 'org.jetbrains.kotlin:1.2.30'
+    plugins "Kotlin"
 }
 ```
 


zolotov:
> An easier way to do this is to simply mention "Kotlin"

It's not an easier way, it's a different way for other purposes.

@JohnHake please consider rewriting the page. I think Kotlin is not the best option for describing external plugins dependencies since it's bundled into IDEA. Php plugin might be a better option.

thsaravana:
@zolotov Oh, so do you mean to say that `"Kotlin"` was possible, because it was bundled with Intellij ? May be you could explain with a different plugin, but please update the docs removing `'org.jetbrains.kotlin:1.2.30'`, or by updating it with a version that actually works.

zolotov:
> so do you mean to say that "Kotlin" was possible, because it was bundled with Intellij ? 

yes

> @JohnHake please consider rewriting the page. I think Kotlin is not the best option for describing external plugins dependencies since it's bundled into IDEA. Php plugin might be a better option.

@JohnHake ping

PR #127, opened 2018-11-04
ToolWindow example does not work with multiple windows

boukeversteegh:
The ToolWindow code example binds the form to the ToolWindowFactory instance, making its content available to only one project window at a time. It leads to content jumping from one window to the other.

Obviously, that is not the behavior any plugin developer would want, so I find the example very confusing. Apparently I'm not alone (see https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000625310-ToolWindow-Instantiation-for-each-project-opened)

I've changed the example minimally to create a separate bound class, to which each instance of the form can bind.


yole:
Thanks for your contribution! Binding a form to a ToolWindowFactory is indeed a completely wrong thing to do, and it's really unfortunate that our example demonstrated that.

PR #129, opened 2018-11-07
Fix Simple Language Plugin for latest IDEA version

shawwn:
Howdy,

I noticed the Simple Language Plugin's Annotator test fails with the latest IDEA SDK:

```
IntelliJ IDEA 2019.1 EAP (Community Edition)
Build #IC-191.SNAPSHOT, built on 10:23, November 7, 2018
JRE: 1.8.0_162-ea-b03 x86_64
JVM: Java HotSpot(TM) 64-Bit Server VM by Oracle Corporation
macOS 10.13.4
```

This PR does the following:

1. Fixes syntax errors in AnnotatorTestData.java

2. Prevents LineMarkerInfo from being registered on non-leaf elements

After these commits, the test suite passes.


YannCebron:
the sample plugin has been revised and this bug is fixed

PR #139, opened 2019-01-16
Added section for the usage of the token with Gradle

halirutan:


plugin_repository/api/plugin_upload.md
```diff
@@ -43,3 +43,13 @@ Curl command example:
 ```
 curl -i --header "Authorization: Bearer perm:qwertyasdfghzxcvb" -F xmlId=ro.redeul.google.go -F file=@Go-0.11.1197.zip -F channel=nightly https://plugins.jetbrains.com/plugin/uploadPlugin
 ```
+
+## Using Gradle with the `gradle-intellij-plugin`
+
+Upload your plugin directly with the `plublishPlugin` task using the token. Create a file `gradle.properties` that you keep local and don't share in your public repository. Add the following content
```
klazuka: typo in "with the `plublishPlugin`"

zolotov: Are you sure that Gradle propagates properties from `gradle.properties` to the plugin's extensions? I don't think it works

halirutan: Hmm, it is the setting I use for some years. Before, I had `username` and `password` in the `gradle.properties` and when you began giving me warnings that I should switch to tokens, I did. It works for me but I'm not even close to an expert with Gradle.

halirutan: Well well.. thanks.

zolotov: could you share the project where it works? or at least content of `build.gradle` file?

I've created the following project and `./gradlew hello` prints `null` despite the fact `publishPlugin.token` is defined in `gradle.properties`.

```
plugins {
      id "org.jetbrains.intellij" version "0.4.1"
}

intellij {
    version '2018.3.2'
}

task hello {
    doLast {
        println publishPlugin.token
    }
}
```

halirutan: Sorry for the confusion, of course you need something like this in your main `build.gradle`

```
// On travis, we don't want to deploy to Jetbrains that requires credentials
if (project.file('gradle.properties').exists()) {
    apply from: "gradle.properties"
}
```

For me, the `hello` task prints the token.

zolotov: Ouch, now I understand why you need quotes in properties file :)

I'd not recommend users doing this. `apply from` treats passed file as a Groovy file and just evaluates it, no magic here.

First of all, it just sounds wrong to evaluate properties file as Groovy. Any properties-like comment (`# hello`), dollar-sign or property value without quotes will break your entire build script. 
You could use any other file, not `gradle.properties` for this, but `myCustomScript.groovy`.

In the second, `gradle.properties` is a special file and in fact it does allow you to substitute some properties, but not in a way you described it.

Here is the documentation about the properties files and how you can use their content in your build script: https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties.

Won't merge it, sorry.

zolotov:
I have no right to reject it. For those who will review this, the instructions are kind of harmful, I'd suggest not add them.

yole:
Closing according to the comments above.

zolotov:
Yes, there is an alternative: use `gradle.properties` file as it was supposed to be used and override values via system properties or via environment variables. This is what described in the link I gave before: https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties.

PR #140, opened 2019-01-19
Show plugin dependencies with Kotlin build script

alshain:
Explains how to set up plugin dependencies with a Kotlin Gradle build script.
Also links to the gradle-intellij-plugin for what values are allowed for plugins/setPlugins.

Moved down instructions on how to add plugin dependencies to IntelliJ SDK to promote Gradle approach.

### DIFF
--- a/basics/plugin_structure/plugin_dependencies.md
+++ b/basics/plugin_structure/plugin_dependencies.md
@@ -5,7 +5,21 @@ title: Plugin Dependencies
 Your plugin may depend on classes from other plugins, either bundled, third-party or your own. In order to express such 
 dependencies (e.g. Kotlin), you need to perform the following steps:
 * If the plugin is not bundled, run the sandbox instance of your target IDE and install the plugin there.
-* Add the jars of the plugin you're depending on to the classpath of your *IntelliJ Platform SDK*.
+* If you are using Gradle with a Kotlin build script to build your plugin, use `setPlugins()`[^gradleplugin] within the `intellij` block, for example:
+
+```kotlin
+intellij {
+        setPlugins("Kotlin", "gradle", "junit")
+}
+```
+* If you're using Gradle with a groovy build script to build your plugin,  add the dependency to the `plugins`[^gradleplugin] parameter of the `intellij` block in your build.gradle, for example:
+
+```groovy
+intellij {
+    plugins 'org.jetbrains.kotlin:1.2.30'
+}
+```
+* If you aren't using Gradle, add the jars of the plugin you're depending on to the classpath of your *IntelliJ Platform SDK*.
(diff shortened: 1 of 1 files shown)


basics/plugin_structure/plugin_dependencies.md
```diff
@@ -5,7 +5,21 @@ title: Plugin Dependencies
 Your plugin may depend on classes from other plugins, either bundled, third-party or your own. In order to express such 
 dependencies (e.g. Kotlin), you need to perform the following steps:
 * If the plugin is not bundled, run the sandbox instance of your target IDE and install the plugin there.
-* Add the jars of the plugin you're depending on to the classpath of your *IntelliJ Platform SDK*.
+* If you are using Gradle with a Kotlin build script to build your plugin, use `setPlugins()`[^gradleplugin] within the `intellij` block, for example:
```
YannCebron: Converting Gradle to Kotlin DSL is a well-known process, it seems somewhat redundant to give this example, additionally.

alshain: Maybe for some, it would have made things smoother for me since I hadn't used the Kotlin DSL before a lot. Also, code completion actually shows the `plugins` field but it couldn't be assigned to.

YannCebron: Ok. The Kotlin DSL sample uses bundled plugins notation, though ("Kotlin", "gradle", "junit"), could you please align this with the Gradle sample below which uses plugin from repository ("org.jetbrains.kotlin:1.2.30")? Thanks!

alshain: @YannCebron Done.
Btw, the example "org.jetbrains.kotlin:1.2.30" doesn't work, I just tried it in a project of mine. Using "org.jetbrains.kotlin:1.3.11-release-IJ2018.3-1" did work though. It seems a lot more convenient to just use "Kotlin" instead.

YannCebron: Could you please fix the version identifier to the working one for both cases then, please? Thanks!

NB using simply "Kotlin" will use the bundled Kotlin version, but not the one from repository (see gradle-intellij readme)

alshain: @YannCebron Done.
Yeah, I understand that, but maybe you could add a paragraph on why would choose one over the other? I'm assuming that referencing a specific version from the repository is better than using the bundled one, otherwise you wouldn't have done the example that way. But considering that for the kotlin plugin, there's different releases per IDE version and per IDE as in IntelliJ/Android Studio etc, it seems much more complicated rather than just using a bundled version of the plugin.

YannCebron:
Thanks a lot for your contribution!

PR #142, opened 2019-01-21
Added the information about NuGet Upload API

LChernigovskaya:
According to https://youtrack.jetbrains.com/issue/MP-1879


yole:
The information on using NuGet should be included in the ReSharper plugins SDK, not here. See https://www.jetbrains.com/help/resharper/sdk/README.html and https://github.com/JetBrains/resharper-devguide

LChernigovskaya:
The opportunity to upload .Net plugins to plugins.jetbrains.com via NuGet API is described here. Please, see https://youtrack.jetbrains.com/issue/MP-873

PR #146, opened 2019-01-28
Add a tip describing when you might want to override `PsiElement#getUseScope`

klazuka:
While digging through [intellij-rust](https://github.com/intellij-rust/intellij-rust) for tips/tricks, I noticed that they overrode `PsiElement#getUseScope` [for function parameters](https://github.com/intellij-rust/intellij-rust/blob/b34cb33882f46b9d57cef2d58beaa4907fcc1e2d/src/main/kotlin/org/rust/lang/core/psi/ext/RsPatBinding.kt#L78). I was pleased to find this as it allowed me to make some nice improvements in my plugin [intellij-elm](https://github.com/klazuka/intellij-elm).

So I figured it would be good to add this to the documentation.


YannCebron:
Thanks!

PR #152, opened 2019-02-09
Fix various configuration commands in the 'Publish plugin with Gradle' tutorial

PHPirates:
Rationale behind all the changes:

* Quotes are removed in the properties file because when present Gradle assumes they are part of the username and password, resulting in (most probably) incorrect credentials being submitted
* `apply from: "/path/to/custom.properties"` does not work anymore, apparently it is [deprecated since Gradle 2](https://stackoverflow.com/a/20639585/4126843). There are multiple ways around it, but this is the one that worked for me (in the sense that IJ recognised whether the variables in the properties file were actually used, for example) and was highest ranked [on this StackOverflow question](https://stackoverflow.com/a/11751301/4126843). I am not saying this has to be the right way to solve it, if you suggest a different way then I can change it.
* I changed the `apply plugin: 'org.jetbrains.intellij'` to using the `plugins` block. As far as I know, using the `plugins` block is preferred now, and at least for me just using the `apply plugin` does not work anymore (Gradle 5).
* The Gradle task `runIdea` does not exist, probably a typo or an old task name.
* The `intellij.publish.channel` option does not exist (anymore), this can be seen in the link to the docs I added at the very end, it is now `publishPlugin.channels`.
* I inserted an example which shows that you can put the `publishPlugin.channels` configuration also in the `publishPlugin` block.

I hope I didn't write too much, but I think these changes aren't trivial and may need to be discussed :) 

PS Excellent tutorial! I am sometimes hard pressed to find good documentation about configuring Gradle tasks (not to mention the Kotlin DSL...), and the IJ docs really help a lot.


tutorials/build_system/deployment.md
```diff
@@ -38,8 +39,20 @@ git update-index --assume-unchanged gradle.properties
 
 If your project already has a custom `gradle.properties` file, you may create a custom `*.properties` file, and load it manually. For example:
 
+```groovy
```
zolotov: I'd get rid of this completely. If you put your properties in the `gradle.properties` that will be automatically imported into `project` properties. See https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties

tutorials/build_system/deployment.md
```diff
@@ -49,7 +62,9 @@ The gradle-intellij-plugin provides a number of [configuration options](https://
  The Gradle plugin will also update the `<idea-version since-build=.../>` values within the `plugin.xml` file to match the `intellij.version`, valid until the last release in the current major version, however you can disable this feature by setting the `intellij.updateSinceUntilBuild` option to `false`.
 
 ```groovy
-apply plugin: 'org.jetbrains.intellij'
+plugins {
+    id 'org.jetbrains.intellij' version '0.4.2'
```
zolotov: there is 0.4.5 now that is able to work with java 11

tutorials/build_system/deployment.md
```diff
@@ -61,7 +76,7 @@ group 'com.jetbrains'
 version '1.2' // Update me!
 ```
 
-When you run `gradle runIdea` with a build script containing the above snippet, Gradle will download the appropriate version of IntelliJ IDEA from either a [Snapshot](https://www.jetbrains.com/intellij-repository/snapshots) (time-based) or [Release](https://www.jetbrains.com/idea/help/managing-plugins.html) (version based) repository, configure the plugin sandbox, install your plugin, and launch a new instance of the IDE. This task can be run directly from the command line, without any prior tooling assistance. 
+When you run `gradle runIde` with a build script containing the above snippet, Gradle will download the appropriate version of IntelliJ IDEA from either a [Snapshot](https://www.jetbrains.com/intellij-repository/snapshots) (time-based) or [Release](https://www.jetbrains.com/idea/help/managing-plugins.html) (version based) repository, configure the plugin sandbox, install your plugin, and launch a new instance of the IDE. This task can be run directly from the command line, without any prior tooling assistance.
```
zolotov: I think link to the releases should be https://www.jetbrains.com/intellij-repository/releases

tutorials/build_system/deployment.md
```diff
@@ -61,7 +76,7 @@ group 'com.jetbrains'
...
+When you run `gradle runIde` with a build script containing the above snippet, Gradle will download the appropriate version of IntelliJ IDEA from either a [Snapshot](https://www.jetbrains.com/intellij-repository/snapshots) (time-based) or [Release](https://www.jetbrains.com/idea/help/managing-plugins.html) (version based) repository, configure the plugin sandbox, install your plugin, and launch a new instance of the IDE. This task can be run directly from the command line, without any prior tooling assistance. 
 
 For best results, plugin developers should build against a fixed version, rather than `LATEST-TRUNK-SNAPSHOT`. For information about available versions of the IntelliJ Platform, you may consult the following URLs for the most recent updates:
```
zolotov: `LATEST-TRUNK-SNAPSHOT` is obsolete, I think the entire paragraph can be removed

tutorials/build_system/deployment.md
PHPirates: Thanks for the review! 
Yes I know, but let me write down my original thought: In this case we are talking about username+password, which you do not want to commit to a VCS. But I think it is quite common that the `gradle.properties` file _is_ committed to a VCS, so you wouldn't want to add it to that file but keep it in a separate `credentials.properties` file, say, which you then don't commit to a VCS. Do you agree? If so, then this idea should maybe replace the "If your project already has a custom gradle.properties file" because I think that would not be a real reason for doing this, for the reason you mentioned.

tutorials/build_system/deployment.md
PHPirates: Good catch, of course no reason to put an old version there, although I assume it is not intended to update this on every new release. Will change it.

tutorials/build_system/deployment.md
PHPirates: Ah I wondered about that, because I vaguely remembered that I had tried it before and it didn't work. That would explain.

tutorials/build_system/deployment.md
zolotov: yes, I think no need to update it on every release. Maybe add a note that it makes sense to check whether the newer version is available at https://plugins.gradle.org/plugin/org.jetbrains.intellij or subscribe to releases at https://github.com/JetBrains/gradle-intellij-plugin/releases?

tutorials/build_system/deployment.md
zolotov: >  Do you agree?

Yes, it's a valid point and valid approach for sure.

I usually set empty username and password in `gradle.properties` and override them on the CI using system property `-Dorg.gradle.project.myPropertyName=username` or via environment variable ORG_GRADLE_PROJECT_myPropertyName.

tutorials/build_system/deployment.md
PHPirates: Yeah I was thinking about that too, I'll do that.

tutorials/build_system/deployment.md
PHPirates: Oh, but that is much better than trying not to commit a certain file! I didn't know that was possible. I tested both ways, the argument and environment and they both work. If you don't mind I'm going to write that down and replace the mentioned paragraph, unless it's already written down elsewhere and I should link to it.

PHPirates: In fact, doing it this way you don't need to put the properties in the `gradle.properties` file, so why use that at all?

zolotov: > In fact, doing it this way you don't need to put the properties in the gradle.properties file, so why use that at all?

To have default values. If you reference these properties in the build script and do not pass anything via environment or argument, you'll get a compilation error.

Yes, please, write it down too.

PHPirates: Ah of course that is correct, forgot that sorry.
I think I have applied all suggestions now, I also resolved the merge conflict which was introduced because Yann noticed the same wrong releases url (ac39402cba623fb071c185051fb69821d219a43c).
Thanks again and no worries about any delay in reviewing.

zolotov:
thanks a lot! merged

PR #153, opened 2019-02-23
Removed boilerplate code from example with lambdas

paxromana96:
Given the advent of Java 8, anonymous classes for one-line commands are more visual distraction than they are helpful.
I replaced the two used here as action listeners for buttons with equivalent lambdas.


JohnHake:
Thank you for contributing.

paxromana96:
Happy to help. Thanks for taking a look at it.

PR #163, opened 2019-04-09
Make intellijPublishUsername and intellijPublishPassword optional

xylo:
The current guide with directly referencing `intellijPublishUsername` and `intellijPublishPassword` is somehow unpractical for plugin contributors that do not want to publish the plugin but expect the plugin project to work out of the box after checkout.  When they try to open the gradle project in IDEA they are faced with error messages like:
```
[...]
> Could not get unknown property 'intellijPublishUsername' for task ':publishPlugin' of type org.jetbrains.intellij.tasks.PublishTask.
[...]
```
Without fixing this error the project cannot be opened.  Users without gradle experience might give up at this point.  Therefore it's better to make these parameters optional.

### DIFF
--- a/tutorials/build_system/deployment.md
+++ b/tutorials/build_system/deployment.md
@@ -24,8 +24,8 @@ Then refer to these values in `publishPlugin` task in your `build.gradle` file:
 
 ```groovy
 publishPlugin {
-    username intellijPublishUsername
-    password intellijPublishPassword
+    username findProperty('intellijPublishUsername')
+    password findProperty('intellijPublishPassword')
 }
 ```
 


zolotov:
The entire task in optional, just remove it if you don’t need publishing.
The properties you made optional have nothing to do with gradle plugin, they are your own properties, you can rename them, remove them or set to empty values.

I’m against merging this, it is quite strange change proposal.

xylo:
Actually, I AM using the publishing task, that's why I would like to keep it.  However, I have lot's of contributors and I want to make contributions as easy as possible.  That's why I did this change for my project.  And I just thought that many plugin developers run into the same issue that contributors claim that they cannot build the project because of missing properties.

Anyway, if you think that this is not an improvement, then just close this PR.  That's fine with me.

zolotov:
> However, I have lot's of contributors and I want to make contributions as easy as possible. 

You can put empty values in `gradle.properties`, as it described in the documentation:

```
To do this, place the following information inside a file called gradle.properties under your project's root directory, or inside GRADLE_HOME/gradle.properties.

intellijPublishUsername=YOUR_USERNAME_HERE
intellijPublishPassword=YOUR_PASSWORD_HERE
```

> PS: Actually, I just wanted to create an issue for this and discuss this, but the intellij-sdk-docs does not allow issues. :(

@YannCebron what do you think about it?

zolotov:
Issues can be created in https://youtrack.jetbrains.com/issues/IJSDK. We promote it in the very first section in the README, not sure whether it can be promoted in a better way.

image

xylo:
> You can put empty values in `gradle.properties`, as it described in the documentation:
> 
> ```
> To do this, place the following information inside a file called gradle.properties under your project's root directory, or inside GRADLE_HOME/gradle.properties.
> 
> intellijPublishUsername=YOUR_USERNAME_HERE
> intellijPublishPassword=YOUR_PASSWORD_HERE
> ```

Yeah, but this is unnecessary effort for contributors, since they actually never need these settings.  Anyway, I think we have just different opinions about how simple the entry point should be.  I will just stick with my solution.  That's fine.

> Issues can be created in https://youtrack.jetbrains.com/issues/IJSDK. We promote it in the very first section in the README, not sure whether it can be promoted in a better way.

Oops. Sorry then.

zolotov:
> Yeah, but this is unnecessary effort for contributors

What is effort exactly? Fetching these two lines from VCS doesn't look like a effort for contributors to me.

xylo:
> What is effort exactly? Fetching these two lines from VCS doesn't look like a effort for contributors to me.

OK, but that would be inconsistent with the following line of the guide:
> If you place a `gradle.properties` file in your project’s root directory, please ensure that this file is ignored by your version control tool.

xylo:
I'm sorry, that the discussion got a bit lengthy and I already feel like becoming a troll. :(
Please note that I only made a suggestion for what I thought is an improvement.  Anyway, this whole issue was only a tiny improvement of the workflow.  So it's probably not worth this issue.  Sorry for taking your time.

zolotov:
If you put the password there, of course, you should not put it in VCS.

The documentation is not supposed to tell you what exactly should be done. Please read it entirely and decide for yourself what approach is better for you. I think introducing things like `findProperty` brings nothing but noise and hide the initial thought that is supposed to be told.

PR #164, opened 2019-04-19
Add references to VFS and Document utils

jonathonblonchek:
Adding information on specific utils we have found helpful and needed to search the codebase to find. Let me know if there is a better place for these things.


yole:
Thanks a lot!

PR #185, opened 2019-07-20
TASK: clarify docs for testing a formatter

skurfuerst:


YannCebron:
Could you please explain why this code would be necessary in context of the Sample Language Plugin tutorial?

skurfuerst:
OK, that's actually not needed :-) I figured out that my plugin was not loaded at all in the testcases, due to missing dependencies of a dependent plugin ;-)

So feel free to close this; or we add the hint about the FormatterTestCase nevertheless.

All the best,
Sebastian

YannCebron:
Thanks, added link

PR #191, opened 2019-08-07
updat quick start doc when version is above 2019.2 needs plugins 'java'

WEIZIBIN:
As Matt Ellis said in post https://blog.jetbrains.com/platform/2019/06/java-functionality-extracted-as-a-plugin/
In the latest EAP build of IntelliJ IDEA 2019.2 we’ve extracted Java functionality into a separate plugin.

### DIFF
--- a/tutorials/build_system/prerequisites.md
+++ b/tutorials/build_system/prerequisites.md
@@ -121,6 +121,11 @@ The New Project Wizard produces the `my_gradle_plugin` project `build.gradle` fi
   
   // See https://github.com/JetBrains/gradle-intellij-plugin/
   intellij {
+      // you need to add plugins 'java' when you version is 2019.2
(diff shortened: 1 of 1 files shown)


YannCebron:
Thanks for your patch, but this applies only to those plugins requiring Java functionality.

We've listed this change on http://www.jetbrains.org/intellij/sdk/docs/reference_guide/api_notable/api_notable_list_2019.html#notable-changes-in-intellij-idea already.

WEIZIBIN:
Thanks.It's hard to find this change.
Is that means If I use 'Java functionality' in my plugin. I have to build two version (one for prior 2019.1, one for above 2019.2) plugin?

YannCebron:
I've updated the start page to link the API change pages more prominently, also put a temporary NOTE about this specific issue. Thanks for your feedback! http://www.jetbrains.org/intellij/sdk/docs/welcome.html

PR #204, opened 2019-09-16
Correct more links

jbeckers:


YannCebron:
Thank you for spotting all these

PR #206, opened 2019-09-18
Change IconLoader to IconManager

patrick-dedication:
This changes IconLoader to IconManager just for the simpleplugin demo as I found it while reading the tutorial.
I'm willing to change this for all occurences of the IconLoader in this rep, just wanted to make sure this is wanted.


yole:
Thanks, but the original code is correct. conLoader is the correct API to use for third-party plugins; IconManager exists only for use in some parts of the platform.

patrick-dedication:
Sorry, just saw the IconLoader is still there, I've looked in the wrong directory for it.

PR #215, opened 2019-10-22
Add info about selected tab customization

mallowigi:


mallowigi:
Allô?

YannCebron:
@mallowigi Why did you close this PR?

mallowigi:
Well it’s pending for a while so I thought maybe you already fixed it on your own

OlyaB:
Selected tab background and underline colors were added to the editor scheme to allow quickly correct tab visibility/contrast for people who have trouble spotting the selected tab (the default may be too subtle for them). This is possible through a custom theme but this seems too much if you need to adjust just the selected tab colors. The ticket about this issue is  https://youtrack.jetbrains.com/issue/IDEA-126775.

I'm not sure we should explain this in custom themes. If you want to change tabs in a custom theme, you can do that in the json file.

Also, adding xml tags like for VCS file status colors is not required, there are UI settings for that:
image

mallowigi:
ok

PR #218, opened 2019-10-25
Document move of FocusDegree to Lookup

nd:
Affects only kotlin usages, in java it is still possible to access moved class as LookupImpl.FocusDegree.


nd:
Decided to roll back the change

PR #222, opened 2019-11-13
IJSDK-676 Python Plugin Dev

JohnHake:
@Paul - Please review products/pycharm.md in particular:
* Are there any blog posts, white papers, or other documents we could mention on this page?
* Are there any open-source PyCharm plugins that would be good examples to mention on this page?

@YannCebron - the old products/pycharm.md page said "The modules are defined in the CommunityRepositoryModules.groovy file.", but this appears to be deprecated. It refers to a newer definition, but it appears to be a class to hold the modules at runtime. (Is there a way to dump the class contents at runtime?)


yole:
If we do create separate sample plugins for each IDE, I think we need to demonstrate using the APIs specific to each IDE in each of these samples. I don't think there's any value in having eight copies of a hello world plugin, with the only difference being the dependencies in build.gradle.

code_samples/product_specific/pycharm_basics/build.gradle
```diff
@@ -0,0 +1,29 @@
+// Copyright 2000-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+
+plugins {
+  id 'java'
+  id 'org.jetbrains.intellij' version '0.4.10'
```
YannCebron: 0.4.12

code_samples/product_specific/pycharm_basics/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,44 @@
+<idea-plugin>
+
```
YannCebron: please reformat file

products/pycharm.md
```diff
@@ -1,19 +1,33 @@
...
 
-### Platform Open API
+## Configuring Plugin Projects Targeting PyCharm
+The configuration of PyCharm plugin projects follows the methods described in [Configuring Plugin Projects using a Product-Specific Attribute](dev_alternate_products.md#configuring-plugin-projects-using-a-product-specific-attribute), and [Configuring the plugin.xml File](dev_alternate_products.md#configuring-the-plugin-pluginxml-file).
+The table below summarizes the attributes to set in the `build.gradle` file:
 
-The APIs from the standard set of Platform Open API modules can be safely used in plugins. Public symbols defined in these modules have a lifecycle, and will be marked as deprecated for several releases before being removed.
+| Gradle Plugin Attribute | Value |
+|-----------|-------|
+| `intellij.type` | `PY` or `PC` |
```
YannCebron: please explain what PY and PC stand for

products/pycharm.md
```diff
@@ -1,19 +1,33 @@
...
 
-### PyCharm Open API
+See the SDK code sample [`pycharm_basics`](https://github.com/JetBrains/intellij-sdk-docs/tree/master/code_samples/product_specific/pycharm_basics/) for an example configuration.
 
-Symbols defined in the **python-psi-api** and **python-openapi** modules are also public API and considered stable.
+## Available PyCharm APIs
+The APIs from the set of [Modules Available in All Products](/basics/getting_started/plugin_compatibility.md#modules-available-in-all-products) can be used in PyCharm plugins.
+APIs for PyCharm Professional are defined in the plugin [Python](https://plugins.jetbrains.com/plugin/631-python).
+The APIs for PyCharm Community are defined in the plugin [Python Community Edition](https://plugins.jetbrains.com/plugin/7322-python-community-edition/).
+These are considered stable APIs, but care should be taken to test your plugin with any version of PyCharm you wish to support.
```
YannCebron: can we list specific packages that are considered "safe"? probably Python-PSI

YannCebron:
Sample plugin: https://github.com/jansorg/pycharm-flake8

products/pycharm.md
JohnHake: Per your comment on IJSDK-648A, I changed "Gradle plugin" to "gradle-intellij-plugin".

I added links from the entries in the "`gradle-intellij-plugin` Attributes" column to the relevant portions of the gradle-intellij-plugin README for each attribute. Seems handy.

I also added "`PY` for PyCharm Professional Edition, or `PC` for PyCharm Community Edition"

products/pycharm.md
JohnHake: As I mentioned in the PR notes above, I'm not sure how to update the reference to `CommunityRepositoryModules.groovy`, which is deprecated. 

The PyCharm Open API was described as "Symbols defined in the **python-psi-api** and **python-openapi** modules are also public API and considered stable." I _assume_ the same is true now, but I'm not sure how these modules relate to the Python or Python Community plugins.

PR #224, opened 2019-11-19
Improve java example

fouad-j:


JohnHake:
These changes have been incorporated as part of the simple_language_plugin Gradle conversion. Thanks for your contribution.

PR #231, opened 2019-12-02
Added Comma to list of IDEs (nt)

holli-holzer:


yole:
Thanks for your PR! I didn't merge it, because Comma is not a JetBrains product, and you've added it to the list of JetBrains IDEs. I've added it to the correct place on the list.

holli-holzer:
Thank you. Comma is a IDE for `Raku` though. It has nothing to do with `Perl`. `Raku` used to be `Perl6`.

yole:
Thanks for noticing! Updated the description

PR #234, opened 2019-12-05
I think it makes sense to mention SwingX

codenameone:
Right now there is very little mention of other capabilities available to developers working in the IDE.


yole:
SwingX is used in a very small number of classes in our code, and it's quite likely that we'll get rid of it altogether at some point. Also, the right way to build plugin UIs that fit cleanly in the IntelliJ UI is to use IntelliJ components. Because of that, I don't think that advertising SwingX to developers would be helpful.

PR #236, opened 2019-12-09
Update VFS changes section due to deprecation

Javaru:
The "How do I get notified when VFS changes?" section indicated to use VirtualFileManager.addVirtualFileListener(). However that method is deprecated. Updated the section with information on using the VirtualFileManager.VFS_CHANGES message bus topic instead.


yole:
Thanks for the detailed write-up!

PR #237, opened 2019-12-16
Document `@NotNull` for parameters of PsiElementVisitor in 2020.1

jreznot:


reference_guide/api_changes/api_changes_list_2020.md
```diff
@@ -73,3 +73,8 @@ Please see [Incompatible API Changes](/reference_guide/api_changes_list.md) on h
 `com.intellij.psi.stubs.PrebuiltStubsProviderBase` class now extends `com.intellij.index.PrebuiltIndexProvider` and inherits its abstract method `getIndexRoot()`
 : Use `com.intellij.psi.stubs.PlatformPrebuiltStubsProviderBase` instead.
 
+`com.intellij.psi.PsiElementVisitor` method `visitElement` parameter marked `@NotNull`.
```
chashnikov: Note that this will break compatibility only with those inheritors, which declare parameter type as nullable. So I would change this to 
> This may break source-compatibility with inheritors written in Kotlin if they declare parameter type as nullable.

chashnikov: Hmm, after I posted my comment I saw that the text was changed, and I cannot find the original text anymore. But I remember that the original text mentioned that only source-compatibility is affected (which looks important for me) and that not only `visitElement` but also all other methods were annotated.

YannCebron: we have the same text on http://www.jetbrains.org/intellij/sdk/docs/reference_guide/api_changes/api_changes_list_2018.html

chashnikov: "Remove ?'s in your Kotlin code" looks obscure for me. Why not mention that only inheritors are affected? Is it so important to make the descriptions as short as possible?

chashnikov: > we have the same text on

Yann, do you agree that such explanations are obscure and should be clarified?

YannCebron: I'm not against changing it, but then we should have the same text everywhere obviously

YannCebron: I will change all such occurrences in docs and also add all other methods of `PsiElementVisitors`

PR #240, opened 2019-12-18
IJSDK-672 DataGrip product page

JohnHake:
Added workaround for error `Cannot load resource: /idea/0xDBEApplicationInfo.xml`.


YannCebron:
LGTM

PR #243, opened 2019-12-25
Added a sample as this was pretty error prone

codenameone:
Also added some clarifying notes all based on this discussion: https://stackoverflow.com/questions/59443789/text-field-with-standard-psielement-auto-completion-in-intellij-plugin


JohnHake:
Thanks for contributing!

PR #244, opened 2019-12-26
Add override annotation in SimpleParserDefinition.java

mongolyy:
I think override annotation isn't required, but writing annotation is better


yole:
Thanks for your contribution!

PR #245, opened 2020-01-15
Notice about annotation migration and possible Kotlin compatibility problem

amaembo:
…roblems


YannCebron:
Thanks, fixed via 6254ca565ff5c08561720bdb99761e9193ecb1e8

PR #247, opened 2020-02-02
Fixed NullPointerException in SimpleLanguage plugin tutorial

schw4rzlicht:
The sample code allows `SimpleUtil.findProperties(Project project, String key)` to return null instead of an empty collection. This leads to `NullPointerException` when `SimpleAnnotator` tries to get the size of the collection.


JohnHake:
Thanks for your contribution. `SimpleUtil.findProperties()` already reflects the following solution as part of other work:
`return result != null ? result : Collections.<SimpleProperty>emptyList();`
However, I think your solution is more elegant. I will incorporate your suggestion in work I am completing this week. Thanks!

PR #249, opened 2020-02-10
`isApplicable` should be used instead of `condition`

develar:


develar:
Maybe we can introduce some kind of version switcher? It is tedious to add note about version. e.g. soon I will open pull request to add diagramm about getting service, but it is also applicable only for a new version.

YannCebron:
we don't have plans to introduce version switcher. the general rule is that SDK docs should document current EAP status, but also last 2-3 major versions in parallel.

develar:
Do we have tech writer who can do this editing?

YannCebron:
will be handled as part of https://youtrack.jetbrains.com/issue/IJSDK-753, thanks for the patch

PR #251, opened 2020-02-16
Fix typo

olada:
While going through the code sample of the simple language plugin demo, I just found a very little typo.


JohnHake:
Thanks for contributing!

PR #252, opened 2020-02-19
Update formatting of Condition type

thepieterdc:
The generic type of the `Condition` class was incorrectly formatted.


YannCebron:
Thanks for spotting!

PR #257, opened 2020-03-02
Add NotificationGroup with example

alexengrig:
Because `Notifications.Bus.register()` is deprecated.


YannCebron:
Thanks. The usual pattern is to create/store `NotificationGroup` simply as static member

`  private final NotificationGroup NOTIFICATION_GROUP = new NotificationGroup("Groovy DSL errors", NotificationDisplayType.BALLOON, true);`

and then call it like so:

`NOTIFICATION_GROUP.createNotification(...)`

Could you please update the code snippets? Thanks

YannCebron:
Thanks for your contribution!

PR #258, opened 2020-03-04
add note about extracted images plugin

develar:
https://youtrack.jetbrains.com/issue/IDEA-234089


YannCebron:
Thanks, I've incorporated your changes with some edits.

PR #275, opened 2020-04-20
Exception while following the annotator tutorial

carolosf:
A dependency at runtime was missing.


JohnHake:
Good catch! 
I found one other place where a change to the project configuration was needed. So I implemented your suggested changes as well. (See f4647bb)
Thanks for your contribution!

PR #278, opened 2020-04-28
Corrected the file name in plugin.xml

ghsatpute:
When run the project as per documentation gives an error 
```
2020-04-28 20:13:40,701 [   5378]  ERROR - Types.impl.FileTypeManagerImpl - Incorrect name specified in <fileType>, should be Simple File, actual Simple file [Plugin: org.example.SimplePlugin] 
```
Fixed this error by correcting the name.


YannCebron:
Thanks for spotting! It is specified correctly in the actual code sample https://github.com/JetBrains/intellij-sdk-docs/blob/master/code_samples/simple_language_plugin/src/main/resources/META-INF/plugin.xml#L35

PR #279, opened 2020-04-29
Clean up language about *IFileElementType*

mikesamuel:
Language about "the IFileElementType interface" may confuse since *IFileElementType* is defined as a Java `class` thus:

```java
/**
 * A superclass for all element types for root AST nodes in a {@link com.intellij.psi.PsiFile}.
 */
public class IFileElementType extends ILazyParseableElementType
```


yole:
Thanks!

PR #285, opened 2020-05-07
Add missing "add-to-group" element

andrewdefilippis:
Pulled from the docs at https://www.jetbrains.org/intellij/sdk/docs/tutorials/action_system/working_with_custom_actions.html#setting-registration-attributes-manually


YannCebron:
Thanks for spotting!

PR #292, opened 2020-05-20
Avoid using Kotlin objects to implement plugin.xml declarations

breandan:
as per @AlexPl292 acejump/acejump#226

### DIFF
--- a/tutorials/kotlin.md
+++ b/tutorials/kotlin.md
@@ -155,9 +155,9 @@ for building forms. Using GUI designer with Kotlin is currently [not supported](
 If a plugin processes Kotlin code (e.g., providing inspections), it needs to add a dependency on the Kotlin plugin (Plugin ID `org.jetbrains.kotlin`) itself.
 Please refer to [Plugin Dependencies](/basics/plugin_structure/plugin_dependencies.md) for more information.
 
-## 6. Advice
+## 6. Caution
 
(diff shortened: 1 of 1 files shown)


tutorials/kotlin.md
```diff
@@ -155,7 +155,11 @@ for building forms. Using GUI designer with Kotlin is currently [not supported](
 If a plugin processes Kotlin code (e.g., providing inspections), it needs to add a dependency on the Kotlin plugin (Plugin ID `org.jetbrains.kotlin`) itself.
 Please refer to [Plugin Dependencies](/basics/plugin_structure/plugin_dependencies.md) for more information.
 
-## 6. Examples
+## 6. Advice
+
+[Objects](https://kotlinlang.org/docs/reference/object-declarations.html) in Kotlin are known to interfere with the way the platform instantiates singletons. For this reason, it is recommended not to use `object`s for any entities that are referenced in the [plugin configuration file](/basics/plugin_structure/plugin_configuration_file.md).
```
yole, line 160: Can we please explain what's actually going on without using the vague "known to interfere" and "recommended not to" wording? When you register an extension, the platform will create a new instance of the class specified in plugin.xml. Therefore, the implementation of the extension needs to be a class, not a Kotlin object. That's a rule, not a recommendation.

PR #294, opened 2020-05-22
[disposer] for review/edit  IJSDK-112

JohnHake:
These documents are based on [3 Google documents](https://drive.google.com/open?id=1AVFMQ_fySh45xt-fLlmutZ5mjH1W9p20): _disposables.md, Android Studio - Fixing Memory Issues.pdf, and step5.md._

Editorial comments about the Google content can be found by searching the review documents for "\> \*\*Note\*\* \*\*-- Editorial --". These comments indicate where some Google content was integrated, and where some was left separate because it might help with your content additions.

PR #298, opened 2020-06-05
Revised disposers documentation based on #294

yole:


JohnHake:
FWIW: the revisions look good to me; they really tightened up the content. Thanks Dmitry.

PR #306, opened 2020-07-16
Update webstorm.md

Gr33nbl00d:
It is not clear that these settings refer to the section intellij inside build.gradle

### DIFF
--- a/products/webstorm.md
+++ b/products/webstorm.md
@@ -10,7 +10,7 @@ Plugin projects for WebStorm can be developed using IntelliJ IDEA with the `grad
 ## Configuring Plugin Projects Targeting WebStorm
 The configuration of WebStorm plugin projects follows the methods described in [Configuring Plugin Projects using the IntelliJ IDEA Product Attribute](dev_alternate_products.md#configuring-plugin-projects-using-the-intellij-idea-product-attribute), and [Configuring the plugin.xml File](dev_alternate_products.md#configuring-pluginxml).
 
-The table below summarizes the `gradle-intellij-plugin` attributes to set in the `build.gradle` file:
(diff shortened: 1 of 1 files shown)


YannCebron:
I guess we should align this on the other IDE specific pages as well

products/webstorm.md
```diff
@@ -10,7 +10,7 @@ Plugin projects for WebStorm can be developed using IntelliJ IDEA with the `grad
 ## Configuring Plugin Projects Targeting WebStorm
 The configuration of WebStorm plugin projects follows the methods described in [Configuring Plugin Projects using the IntelliJ IDEA Product Attribute](dev_alternate_products.md#configuring-plugin-projects-using-the-intellij-idea-product-attribute), and [Configuring the plugin.xml File](dev_alternate_products.md#configuring-pluginxml).
 
-The table below summarizes the `gradle-intellij-plugin` attributes to set in the `build.gradle` file:
+The table below summarizes the `gradle-intellij-plugin` attributes to set in the `build.gradle` file inside section `intellij`:
```
JohnHake, line 13: The difficulty with this specific change is that not all attributes in the table belong to the Setup DSL `intellij {...}` group. At least one attribute belongs to the Running DSL `runIde {...}` group. 

As an alternative, would it help to link to [Configuring build.gradle using the IntelliJ IDEA Product Attribute](https://jetbrains.org/intellij/sdk/docs/products/dev_alternate_products.html#configuring-buildgradle-using-the-intellij-idea-product-attribute), which has example `intellij {}` and `runIde {}` sections demonstrated for PhpStorm, a similar type of configuration?

JohnHake:
I updated all the product pages with changes similar to your suggestion. Thanks for the contribution!

PR #307, opened 2020-07-20
Include the sourceSet rune for kotlin gradle

nthykier:
Include to necessary rune for kotlin based gradle files to assist people starting from the JetBrains/intellij-platform-plugin-template repo, which uses `build.gradle.kts` by default.

Related issue: JetBrains/intellij-platform-plugin-template#17


hsz:
Thanks!

PR #323, opened 2020-10-21
Add information about breaking changes in SpellChecker

TanVD:
`com.intellij.spellchecker.tokenizer.SpellcheckingStrategy.getDefaultRegularFixes(boolean, String, @Nullable PsiElement)` method removed


hsz:
Thanks @TanVD !

PR #324, opened 2020-10-21
Information about incompatible spellchecker changes

TanVD:
Add information about removal of `com.intellij.spellchecker.quickfixes.ChangeTo.<init>(java.lang.String)`


hsz:
Thanks again!

PR #325, opened 2020-10-23
IJSDK-771 PhpTypeProvider4

wbars:
- Provide actual interface
- Replace old convoluted example with more simple one
- Remove completion coverage, this is already covered in platform documentation


hsz:
Thanks Kirill!

PR #326, opened 2020-10-25
docs: update getIcon usage

zd-zero:
### Motivation

As I was browsing the [documentation](https://jetbrains.org/intellij/sdk/docs/reference_guide/work_with_icons_and_images.html#how-to-organize-and-how-to-use-icons), the `com.intellij.openapi.util.IconLoader#getIcon(@NonNls @NotNull String path)` would now be deprecated and it would be better to show the recommended method to be used which is `com.intellij.openapi.util.IconLoader#getIcon(@NotNull String path, @NotNull Class<?> aClass)` that uses a class for the 2nd parameter. So newer code would not encounter deprecations when they target the new platform build.

### DIFF
--- a/reference_guide/work_with_icons_and_images.md
+++ b/reference_guide/work_with_icons_and_images.md
@@ -26,9 +26,9 @@ Then define a class/interface in a top-level package called `icons` holding icon
 package icons;
 
 public interface DemoPluginIcons {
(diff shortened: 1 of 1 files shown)


hsz:
Thanks Zaerald!

PR #327, opened 2020-10-25
Add yet another example of a plugin

kirillgla:
This might be helpful as an example of a ReSharper plugin that evolved to support Rider as well


hsz:
Thanks Kirill!

PR #330, opened 2020-11-21
Fix DemoRunConfigurationType id naming

Nexen23:
ConfigurationType.getId() javadoc says "The ID of the configuration type. Should be camel-cased without dashes, underscores, spaces and quotation marks."


hsz:
Thanks, Alex - good catch!

hsz:
Thanks, Alex - good catch!

PR #331, opened 2020-11-24
Declare removed method

jonnyzzz:
This is a ```@TestOnly``` method used in Scala. We will update the Scala plugin to remove that use.


YannCebron:
please add

> add a new line with the problem pattern followed by a 2nd line with ": "-prefixed human-readable description and recommended fix/action.

jonnyzzz:
Add, sorry I missed that part in the documentation

PR #333, opened 2020-12-01
remove psiImplUtilClass from Grammar and Parser page

knoxfighter:
Remove the definition of psiImplUtilClass in the Grammar and Parser page. With this file added, it is not possible to successfully compile the Sourcecode used for the Tutorial at this point. Later on in chapter 6.2 this line gets added, also with a link back to chapter 3.3.
This is also something that confused me, while reading the tutorial the first time.


YannCebron:
Thanks for spotting!

PR #341, opened 2020-12-11
Add a note on how to override services

kirillgla:


yole:
I'm not sure we want to document this. It's OK to override platform services in products based on the platform, but we really don't want arbitrary plugins to override platform services.

kirillgla:
OK, good to know. Let's not document it, then

PR #345, opened 2021-01-04
correct link for dynamic plugin listener

Trias:


hsz:
Good catch! Thank you, @Trias.

PR #346, opened 2021-01-04
Document removal of Swing Inspector

yole:


YannCebron:
Thanks, both 'property' and 'resource-bundle' should be enclosed with single quotes (already fixed)

PR #348, opened 2021-01-15
Messaging Infrastructure is more akin to "Publisher Subscriber" pattern than "Observer" pattern

mr3y-the-programmer:
I came across [Messaging Infrastructure doc](https://plugins.jetbrains.com/docs/intellij/messaging-infrastructure.html) and noticed that Messaging Infrastructure is explicitly considered an implementation of _Observer_ pattern, I think it is more convenient to be an implementation of _Publisher Subscriber_ pattern, I know they may look similar but still there are some nuances between them and Thus different usecases for each pattern


YannCebron:
Thanks for your contribution!

PR #354, opened 2021-02-10
add 2021.1 branch number

mbolotov:


YannCebron:
2021.1 branch will be created only when first Beta release is there

mbolotov:
so when a plugin developer adding an EAP build version into the `until-build` section, where they should get this number from?

YannCebron:
branch number != build number for target platform

all exact build numbers for each EAP release are available https://www.jetbrains.com/idea/nextversion/ or by browsing repository https://www.jetbrains.com/intellij-repository/snapshots/

PR #393, opened 2021-03-20
Update run configuration tutorial with newer APIs

takanuva15:
The run configuration tutorial currently shows the usage of very old interfaces even though much newer, cleaner interfaces are now available in the IntelliJ API. I updated the code samples to show the usage of the new APIs and left a note in there about the "original" way to do it by implementing `ConfigurationType`.

I have also updated the code samples to use one of the newer APIs and tested that it's working correctly:
image

I noticed that there is pretty much no documentation for the classes in `runConfigurationType.kt`. Do we want to add something there? (I can write up another gist for that if so)

### DIFF
--- a/code_samples/run_configuration/src/main/java/org/jetbrains/sdk/runConfiguration/DemoRunConfigurationType.java
+++ b/code_samples/run_configuration/src/main/java/org/jetbrains/sdk/runConfiguration/DemoRunConfigurationType.java
@@ -2,40 +2,17 @@
 
 package org.jetbrains.sdk.runConfiguration;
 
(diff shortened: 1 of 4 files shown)


YannCebron:
Thanks, will be incorporated via "global" issue https://youtrack.jetbrains.com/issue/IJSDK-95

FW4248 (approved):

Wanlc888 (approved):

karollewandowski:
Thank you for the PR, but the Run Configuration docs were massively reworked, and your PR became obsolete (see https://github.com/JetBrains/intellij-sdk-docs/pull/1065). All the changes are covered.

PR #414, opened 2021-06-15
[go] add information about missing GoSdkChooserCombo#getSdk

artspb:


serejke:
@YannCebron I've added support for `XXX method moved to the superclass` documentation pattern in the Plugin Verifier.
Could you please update the MD FAQ headers as well?

PR #516, opened 2021-08-03
Document incompatibilities introduced after refactoring

traff:
Document incompatibilities introduced after platform.ide and platform.ide.impl refactoring.

During this refactoring new modules platform.ide.core and platform.ide.core.impl were extracted. The new core modules are UI independent. Unfortunately, the compatibility couldn't be preserved in some cases.


hsz:
Thanks, Dmitry!

PR #520, opened 2021-08-09
Update project.ihp

rayshade:
This will keep files located on the same level as project.ihp available in build because we've recently updated help builder so that it won't automatically touch these files (some other projects are storing notes and remarks with exact file names as their topic files in this root, so we need to differentiate.)


hsz:
Thanks!

PR #522, opened 2021-08-19
Fix single typo in 'forgot'

avan1235:


YannCebron:
`forgo` is correct here indeed

PR #556, opened 2021-10-11
Fix a code error in Settings Tutorial

whitescent:


Nthily:
But storages are actually an `Array` type,

btw, if i write it this way, this error will occur

```
storages = @Storage("SdkSettingsPlugin.xml")
```

``` 
An annotation can't be used as the annotations argument
```

PR #586, opened 2021-11-07
The cert and key command line should be `-file`

Shoothzj:


Shoothzj:
@YannCebron PTAL

hsz:
Thank you!

PR #614, opened 2021-11-30
Fix typos in language_injection.md

Soarex16:


YannCebron:
The page is correct, the extension point name is lowercase, but the interface to implement for this extension point is uppercase. I've just pushed a small update to clarify this better: 95f7f999c7b37fea7967a81e292233b44122fe22

Thanks for your feedback!

PR #671, opened 2022-02-19
Add the cliPath to plugin_signing.md

obiscr:
The latest version must set cliPath to successfully sign, otherwise there will be an exception prompt: getHeaderField("Location") must not be null

Signed-off-by: piercebrands <piercebrands@gmail.com>


obiscr:
According to the test results, it may be caused by network instability.
Hope it gets resolved in the future! Thanks!

PR #680, opened 2022-03-16
Feature documentation (aka QuickDoc in the Features section)

halirutan:
A small section as a general overview for `DocumenationProvider` that explains some high-level stuff and refers to the article in the Custom Language Support section.

### DIFF
--- a/topics/tutorials/code_documentation.md
+++ b/topics/tutorials/code_documentation.md
@@ -3,7 +3,7 @@
 <!-- Copyright 2000-2022 JetBrains s.r.o. and other contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file. -->
 
 [Quick Documentation](https://www.jetbrains.com/help/idea/viewing-reference-information.html#inline-quick-documentation)
-helps users by showing documentation for symbols like method calls inside the editor.
+helps users by showing documentation, e.g., for classes, functions, or methods inside the editor.
 Plugin authors implement
 [`DocumentationProvider`](upsource:///platform/analysis-api/src/com/intellij/lang/documentation/DocumentationProvider.java)
 to show documentation for particular Psi elements.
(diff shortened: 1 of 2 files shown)


topics/tutorials/code_documentation.md
```diff
@@ -0,0 +1,33 @@
...
+
+Several documentation providers can co-exist for the same language, and plugin authors can implement additional ones even for languages not under
+their control.
+If there is more than one provider for the same element, then the first one that returns a value different from `null` wins.
+Although discouraged, the ordering of documentation providers can be influenced by using the `order` attribute when registering the extension.
+For instance, [`python-core-common.xml`](upsource:///python/src/META-INF/python-core-common.xml) uses the following to call the external documentation
+provider before the default one:
+
+```xml
+<lang.documentationProvider
```
karollewandowski, line 25: I like this formatting style (except closing token in a new line) and use it personally, but currently, we align attributes to the first one which is placed in the same line as the element name. I would keep it consistent and would propose changing XML snippets formatting in all docs as this style is more likely to fit in the code snippets widget.

It's minor though, feel free to ignore it.

topics/tutorials/code_documentation.md
```diff
@@ -0,0 +1,33 @@
...
+  language="Python"
+  implementationClass="com.jetbrains.python.documentation.PythonExternalDocumentationProvider"
+  order="before pythonDocumentationProvider"
+/>
```
karollewandowski, line 29: I wouldn't place it in the new line.

#minor

topics/tutorials/code_documentation.md
```diff
@@ -0,0 +1,33 @@
...
+<!-- Copyright 2000-2022 JetBrains s.r.o. and other contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file. -->
+
+[Quick Documentation](https://www.jetbrains.com/help/idea/viewing-reference-information.html#inline-quick-documentation)
+helps users by showing documentation, e.g., for classes, functions, or methods inside the editor.
+Plugin authors implement
+[`DocumentationProvider`](upsource:///platform/analysis-api/src/com/intellij/lang/documentation/DocumentationProvider.java)
+to show documentation for particular Psi elements.
+
+`DocumentationProvider` can be used for existing languages as well as for a custom language implementation.
+For an existing language, the implementation of `DocumentationProvider` is registered as `com.intellij.documentationProvider` extension point (EP),
```
karollewandowski, line 12: This difference is unclear to me. The questions that came to my mind:

- Can I register documentation provider via `com.intellij.documentationProvider` for my custom language?
- Can I register documentation provider via `com.intellij.lang.documentationProvider` for Java (existing language)?

If both answers are "yes", then this paragraph is a bit misleading IMHO.

Isn't it about performance?
So if we know the documentation language, we should register it via `com.intellij.lang.documentationProvider`. If our documentation provider can provide doc in various contexts, then we should use `com.intellij.documentationProvider`. Is it correct or I'm missing the point?

karollewandowski (changes requested):

topics/tutorials/code_documentation.md
YannCebron: 1) you should use `lang` when targeting specific Language only
2) yes, it's possible to augment/override/.. existing DocProvider

topics/tutorials/code_documentation.md
```diff
@@ -3,13 +3,31 @@
...
+helps users by showing documentation, e.g., for classes, functions, or methods inside the editor.
 Plugin authors implement
 [`DocumentationProvider`](upsource:///platform/analysis-api/src/com/intellij/lang/documentation/DocumentationProvider.java)
 to show documentation for particular Psi elements.
 
 `DocumentationProvider` can be used for existing languages as well as for a custom language implementation.
 For an existing language, the implementation of `DocumentationProvider` is registered as `com.intellij.documentationProvider` extension point (EP),
 while custom languages should register it as `com.intellij.lang.documentationProvider` EP.
-For details on how to implement a `DocumentationProvider`, please refer to the [Custom Language Support](documentation_provider.md)
+The difference between both is that the latter requires the `language` attribute when registering the EP in the `plugin.xml` and is only called
```
YannCebron, line 14: use `<path>` for filename

topics/tutorials/code_documentation.md
halirutan: I'm not sure how important it is to be consistent throughout all the docs, but for cases like this one, I really like that style. 2 spaces of indent to make as much of the code visible without scrolling.

topics/tutorials/code_documentation.md
halirutan: Oh boy, when will I finally remember that one? Every. Single. Time.

YannCebron (approved):

PR #684, opened 2022-03-23
Migrate gradle build scripts from groovy to kotlin

karollewandowski:
Migration done without any changes to the properties values.

Unfortunately, Git detects file deletion and addition instead of renaming.


YannCebron:
why did code_samples.yml not run in _Checks_ for this PR??

YannCebron (changes requested):
no verification of PR on GH actions?!

topics/tutorials/kotlin.md
```diff
@@ -86,13 +86,13 @@ Plugins using the [Gradle Build System](gradle_build_system.md) use the [Kotlin
...
-```groovy
+```kotlin
 ```
-{src="kotlin_demo/build.gradle"}
+{src="kotlin_demo/build.gradle.kts" lines="2-"}
 
 ### Use Kotlin for Gradle Build Scripts
 
-Gradle also supports using Kotlin in build scripts: <path>build.gradle.kts</path>.
+Gradle also supports using Kotlin in the build scripts: <path>build.gradle.kts</path>.
```
halirutan, line 95: Remove "the" again. Plural nouns _can_ be used with a definite article but they don't need it. The original version reads better.

code_samples/simple_language_plugin/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  patchPluginXml {
+    version.set("${project.version}")
+    sinceBuild.set("211")
+    untilBuild.set("213.*")
+  }
+
+  test {
+    // This path value is machine-specific placeholder text.
```
halirutan, line 50: ... is _a_ machine-specific...

YannCebron (approved):

PR #772, opened 2022-05-09
welcome.md: contributing to platform

SamB:
Make it easier to find the docs on contributing to the actual software.

I may have overdone it a bit, but I think it's at least accurate.

I write "mostly licensed under Apache 2.0" because there are at least some files under the Eclipse Public License in there -- mostly PyDev, but maybe a few others.

I would probably have linked to https://www.jetbrains.com/legal/community-bundled-plugins/ if I had found it before committing. (I still might.)


YannCebron:
Thanks, you're right. Added link in d2d9b5433a9538a42ac756462e6cafdf02d1872b

PR #777, opened 2022-05-20
Document API breakages in Database Tools and SQL Plugin

kornilova203:
Both changes break only one small plugin https://plugins.jetbrains.com/plugin/13886-copy-as-sql
As already mentioned in #775 I plan to write to plugin's author


YannCebron:
Thanks!

PR #782, opened 2022-05-24
`com.jetbrains.php.docTagParserExtension` EP marked as internal in `php_extension_point_list.md`

i582:


YannCebron:
Thanks, that's actually a discrepancy between the EP declaration `PhpDocTagParserEP` having internal status vs. EP implementation `PhpDocTagParser` status (non-internal)

i582:
Yep, that looks weird too.

YannCebron:
@wbars please review, do not apply patch. we generate these lists from actual sources. thanks

YannCebron:
ApiStatus was updated in sources, changes will be in effect once EP page is regenerated from corresponding branch on release

PR #783, opened 2022-05-24
fixed tags for several EPs

i582:
- `com.jetbrains.php.keyTypeProvider` is actually internal
- `com.jetbrains.php.noReturnProvider` is actually not experimental


i582:
Cool, thank you

PR #795, opened 2022-06-15
added small section to PHP Open API describing `PhpCompletionUtil` utility class

i582:


wbars:
Thank you for the contribution. I'd say in this current form proposed documentation doesn't add too much semantic for the existing API signature, so I'd shelve this for now.

PR #800, opened 2022-06-28
Document api change for VcsLogHighlighter. 2022

a-krasilnikov:
The method signature has been changed. One more parameter added.


YannCebron:
please remove `.` from first line and add it to end of 2nd line, please :)

PR #801, opened 2022-06-30
Use the correct wording

Konafets:
`SimpleFormattingModelBuilder` implements `FormattingModelBuilder` and does not subclassing it.


karollewandowski:
Thank you, Stefano.

PR #806, opened 2022-07-07
Minimize png images

i582:
Before: 27.8mb, after: 20.5mb (-26%)

Minimized by `pngquant`.


YannCebron:
Interesting. I'm a bit sceptical, since I already used PNG optimization via IDE plugin (https://plugins.jetbrains.com/plugin/7942-png-optimizer) which seems also to be a "safe" way to ensure new screenshots are always optimized. It seems this tool needs manual invocation and thus would be harder to "maintain" its results in future.

i582:
> Interesting. I'm a bit sceptical, since I already used PNG optimization via IDE plugin (https://plugins.jetbrains.com/plugin/7942-png-optimizer) which seems also to be a "safe" way to ensure new screenshots are always optimized. It seems this tool needs manual invocation and thus would be harder to "maintain" its results in future.

@YannCebron in general, everything is so, I agree. However, this can be a way to reduce the size right here, and all the rest to be checked by the plugin in the future. Most likely for these optimized images, the plugin will also say OK.

halirutan:
My problem is that `pngquant` is a _lossy_ compression. It actually changes the image contents and similar to jpg compression, artifacts will happen on sharp edges. Most/all our image content is exactly that: GUI and text content with sharp edges. If you visualize the differences between images, you see where it happens (in red):

diff

Even if the differences are not immediately visible when just looking at the processed images, I'd vote for keeping the original images because they are as crisp and clear as they come from the screenshot. Additionally, when I profile the SDK docs page with [one of the biggest images](https://plugins.jetbrains.com/docs/intellij/dev-alternate-products.html#configuring-plugin-projects-using-the-intellij-idea-product-attribute), the stats look like this

image

So even if we scrape off 30% from the image size, the impact will not even be noticeable in the light of all the scripts that run.
Therefore, my 3 points are:

- `pngquant` will potentially reduce the quality of the screenshots
- it's an additional manual step that needs to be done on new images
- it seems to have only a minimal impact on the loading speed of the pages

i582:
@halirutan thanks for such detailed review, I think you are right, the impact from this isn't so visible, so I will close the PR.

PR #854, opened 2022-08-19
Add small section about Spell Checker in IDE Infrastructure

i582:


topics/basics/ide_infrastructure.md
```diff
@@ -111,6 +111,13 @@ See also [](plugin_components.md#application-startup) and [](plugin_components.m
 
 Register [`ApplicationActivationListener`](%gh-ic%/platform/ide-core/src/com/intellij/openapi/application/ApplicationActivationListener.java) [listener](plugin_listeners.md) to be notified of "application focused/unfocused" events.
 
+## Spell Checker
+
+Spell checking dictionaries can be modified using
+[`SpellCheckerManager`](%gh-ic%/spellchecker/src/com/intellij/spellchecker/SpellCheckerManager.java).
```
TanVD, line 117: Instead of SpellcheckerManager you should use one of the extension points here https://github.com/JetBrains/intellij-community/blob/0e2aa4030ee763c9b0c828f0b5119f4cdcc66f35/spellchecker/src/META-INF/SpellCheckerPlugin.xml#L6

TanVD (changes requested):

topics/basics/ide_infrastructure.md
i582: It seems #856 covers this plugin. Also I think this way is more specific for adding individual words to the dictionary. Perhaps a link to #865 should be added here.

YannCebron: what do you mean with "adding individual words to the dictionary" and how is it different from providing this wordlist from dedicated EP?

YannCebron: @i582 any update?

PR #855, opened 2022-08-20
Update API changes in 222.2 for Webpack

vepanimas:
Because of new [support](https://youtrack.jetbrains.com/issue/WEB-55332/Vite-aliases-in-viteconfig-support) for Vite bundler, an existing Webpack API was significantly reworked.

When I worked on those changes, there were no usages, and a tremendous amount of work is already spent on it, so there is no possible way to revert it. Also, it looks like a single plugin that depends on that API uses a `@TestOnly` method (`com.intellij.lang.javascript.buildTools.webpack.WebPackConfigManager.setConfig(WebPackConfig)`), and it's not actually expected.

Also, note that in 223 Webpack and Vite will be moved entirely to separate plugins.


YannCebron:
Thanks!

PR #856, opened 2022-08-22
Add article and tutorial about Spell Checking to the Custom Language Support topic

i582:
Added two articles:

- Reference part describing general information about spell checking.
- Implementation part with simple spell checking strategy.


halirutan:
Nicely done. I'd like to give the text some more love. There are some minor mistakes, and I think we can improve the text flow quite a bit. Especially the distinction between the tokenizer and the splitter is not clear enough for future readers: If the tokenizer "splits elements into words", why is there a text-splitter that splits "text into words"? 😉 

However, you already did the heavy lifting which is the most crucial part. Let me get back to you tomorrow with a plan.

halirutan:
@i582 Sorry for making this "coming back to you tomorrow" a whole week. Would it be OK for you if I worked on your documentation and sent you a good old patch file? I could make some deeper changes in your text and you could look at the diff, patch your files and commit the changes.

The other option would be that you give me commit access to your SDK fork so that I can commit directly. I'm fine with either solution.

halirutan:
While working on the text, I'll leave some comments for discussion here. 

> `SpellcheckingStrategy` implements a strategy for tokenizing the text of PSI elements, which will then be checked for spelling.

I think we need to communicate several things to the reader. The `getTokenizer` method basically selects which "tokenizer" is used for a specific PSI element. However, its `tokenize()` method doesn't break text into words. This is done by providing a text-splitter. The `tokenize()` method lets you make checks on the element and provide the text/range inside the PSI element that will then be fed into the splitter.

I think the method naming in the API is weird and a bit misleading. You can easily see this in your own example in the Simple Language. Check out this screenshot

image

The comment tokenizer should have adjusted the range to get rid of the preceding `#`. So right now, none of the first words in a comment are checked for spelling. Secondly, the values of the key/value pairs are not checked and it only works for the keys because `PlainTextSplitter` does such an awesome job of being intelligent and splitting up the camel-case :)

My suggestion would be to fix both. Create a custom comment tokenizer that adjusts the range in the element to exclude the leading `#` before consuming the token. And for the `SimplePropertyTokenizer`, make a condition to check `PsiElement(SimpleTokenType.KEY)` and `PsiElement(SimpleTokenType.VALUE)` separately. Right now, the tokenizer consumes all of `apikey=ph342m91337h4xX0r5k!11Zz!` and hopes for the best that the splitter sorts it out.

halirutan:
I worked on the implementation. Most notably we now have

- comments will be spell-checked after the initial `#` and whitespace chars
- keys will be checked separately from values without the `=`
- For keys, `Alt+Enter` on errors will open "rename refactoring" instead of just "change to"

image

PR #862, opened 2022-09-26
Gradle/DevKit approach or Plugin/Theme topic content reorganization

karollewandowski:
Unfortunately, the PR is large as the change of structure required adjustments in content to make sense, remove duplications, etc.
The recommended way of reviewing:
- Check the commits - I tried to make them granular enough to understand the changes, and I added reasons for changes when it may be controversial (e.g., replacing "UI Theme" with "Theme") or unclear
- Check the content on the rendered webiste locally: "Part I - Plugins" and "Part X - Themes" to see if they read good and are consistent

I also took the chance and renamed some files related to the changes for better SEO and and moved them for better organization in the project structure.


YannCebron (approved):
as discussed

PR #871, opened 2022-10-28
Android Studio plugin documentation - runIde should be inside tasks block

thsaravana:
The `runIde` has to be inside the `tasks` block to work properly. This info was not captured in the documentation. So added that now.

### DIFF
--- a/topics/products/androidstudio/android_studio.md
+++ b/topics/products/androidstudio/android_studio.md
@@ -76,10 +76,12 @@ intellij {
   plugins.set(listOf("android"))
 }
 
-runIde {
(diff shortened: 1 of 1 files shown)


karollewandowski:
Thanks a lot for the PR.

For Kotlin, indeed, this is a bug and should be fixed.
For Groovy, it's fine. Tasks usually are configured at the top level of the build script, not inside tasks. Could you please revert the change for Groovy?

thsaravana:
I instantly assumed it's the same for Groovy. My bad. Reverted the changes for Groovy.

karollewandowski (approved):

PR #872, opened 2022-10-31
Update API changes in 222.3 for Webpack 

vepanimas:
Webpack is extracted from the JS core and is a separate plugin now.


karollewandowski (approved):
Thanks for the PR.
I'm not sure if we should include `Companion` as well, but it will be reduced in case we should not.

PR #875, opened 2022-11-04
Specify breaking change about UrlPathInlayAction for 2022.3

dtsaryov:
Previously I described the change for 2023.1 by mistake


dtsaryov:
@YannCebron could you review this change please?

PR #876, opened 2022-11-04
Document breaking changes in 2022.3

jreznot:


YannCebron:
1. please revert the change in code sample, they don't target latest GA
2. please remove code inline marker around "`YAMLParserDefinition`" as it is not supported change format
3. change for "com.intellij.codeInsight.template.TemplateContextType.EP_NAME", please link to page which has all details instead

jreznot:
> please link to page which has all details instead

Sorry, I don't get how it must look like

PR #986, opened 2023-02-18
Description of anchor attribute after changed

astrapi69:
The description of the anchor attribute after fixed


karollewandowski:
Thank you for catching and providng a PR.

PR #1086, opened 2023-07-06
Update plugin_services.md (fix missing implements expression)

vidgeus:


karollewandowski:
Thank you very much for catching it and for the fix.

PR #1093, opened 2023-07-27
code_samples/project_view_pane: Cleanup

karollewandowski:
- use more proper APIs
- fix a bug with not refreshing "svg" files and add supporting "jpeg" extension
- fix not working "Folders Always on Top" switch
- fix extracting file extension
- change disposable from `Project` to `ProjectViewPane`
- use `BulkFileListener` instead of `VirtualFileListener` with `Alarm` for updating the tree for immediate updates
- use `Application.invokeLater` instead of `SwingUtilities.invokeLater`
- code cleanup

### DIFF
--- a/code_samples/project_view_pane/src/main/java/org/intellij/sdk/view/pane/ImagesProjectNode.java
+++ b/code_samples/project_view_pane/src/main/java/org/intellij/sdk/view/pane/ImagesProjectNode.java
@@ -14,6 +14,8 @@ import com.intellij.openapi.application.ReadAction;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectUtil;
+import com.intellij.openapi.roots.FileIndex;
+import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.VirtualFileManager;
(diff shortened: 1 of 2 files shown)


code_samples/project_view_pane/src/main/java/org/intellij/sdk/view/pane/ImagesProjectNode.java
```diff
@@ -75,15 +79,51 @@ private Set<VirtualFile> getImagesFiles(Project project) {
...
+              String extension = file.getExtension();
+              if (extension != null && SUPPORTED_IMAGE_EXTENSIONS.contains(extension)) {
+                hasAnyImageUpdate = true;
+                break;
+              }
+            }
+            if (hasAnyImageUpdate) {
+              getImagesFiles(project).clear();
+              scanImages(project);
+              ApplicationManager.getApplication().invokeLater(() ->
```
YannCebron, line 102: use `com.intellij.util.ui.update.MergingUpdateQueue` to prevent too many/"overlapping" updates in UI

YannCebron: use `invokeLater` with `Project.getDisposed` condition overload

YannCebron (changes requested):

code_samples/project_view_pane/src/main/java/org/intellij/sdk/view/pane/ImagesProjectNode.java
karollewandowski: > use `com.intellij.util.ui.update.MergingUpdateQueue` to prevent too many/"overlapping" updates in UI

I think it should not be a problem, as we use `BulkFileListener`, and UI update happens once for bulk if there is at least one image change. Or am I missing something?

YannCebron (changes requested):

code_samples/project_view_pane/src/main/java/org/intellij/sdk/view/pane/ImagesProjectNode.java
karollewandowski: You were right about using `MergingUpdateQueue`. Actually, even though I used `BulkFileListener`, bulks were unexpectedly for me created for each file and UI was updated multiple times. Thanks.

The rest of the issues have been also fixed.

YannCebron (approved):

PR #1097, opened 2023-08-16
Update api_changes_list_2023.md

HackerMadCat:
Documented breaking changes during External System Run Configuration fragment builders refactoring.

This refactoring caused by changes in text completion collecting process. It become async.

https://youtrack.jetbrains.com/issue/IDEA-320168


YannCebron:
You must use the exact patterns from the comment on top of this page. If none match, then start entry without `code` but use plaintext

PR #1143, opened 2023-11-03
Document plugin manager api changes in 2023.3

kesarevs:
Enum values `com.intellij.ide.plugins.enums.PluginsGroupType.FEATURED` and `com.intellij.ide.plugins.newui.SearchWords.ORGANIZATION` were replaced in 2023.3 with `com.intellij.ide.plugins.enums.PluginsGroupType.STAFF_PICKS` and `com.intellij.ide.plugins.newui.SearchWords.VENDOR` respectively.


karollewandowski (approved):
Thanks. The previous version "enum value renamed" was more correct, but we don't support it yet. We will track and fix it on our own.

PR #1144, opened 2023-11-04
sourceCompatibility version should match the version in the code sample

Cazka:


karollewandowski:
Thank you for the fix.

PR #1146, opened 2023-11-10
Update api_changes_list_2023.md (fix typo, add missing method name)

kornilova203:
Noticed that I accidentally omitted method name last time


karollewandowski:
Thanks for the fix.

PR #1147, opened 2023-11-13
Document some incompatible Kotlin plugin changes for 2023.3

darthorimar:
`org.jetbrains.kotlin.idea.compiler.configuration.KotlinIdePluginVersion` and `org.jetbrains.kotlin.idea.compiler.configuration.KotlinIdePluginVersion.Companion` internal API is removed.


YannCebron:
Please remove the wording about "Internal API" as this related to a specific concept in IJ Platform https://plugins.jetbrains.com/docs/intellij/api-internal.html

PR #1157, opened 2023-12-02
Added a file Download section in jcef.md

timKraeuter:
I added a **file download section** in the JCEF part of the documentation.

It would be nice to have a short section pointing out how file downloads can be implemented since they do **not** work out of the box.

I implemented the file download using this method in my plugin ([visual debugger](https://github.com/timKraeuter/VisualDebugger)).

The link I added probably needs some adjustments to fit the other links in the documentation.

Let me know what you think.

### DIFF
--- a/topics/reference_guide/jcef.md
+++ b/topics/reference_guide/jcef.md
@@ -78,6 +78,35 @@ browser.openDevtools();
 
 See [`JBCefTestHelper`](%gh-ic%/platform/platform-tests/testSrc/com/intellij/ui/jcef/JBCefTestHelper.java) and tests in that package.
 
+## File Download
+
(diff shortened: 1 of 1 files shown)


karollewandowski:
Hi Tim. Thank you for the PR, but we will pass on merging it.

The page was almost fully rewritten recently and is in an internal review (it should be merged this week).
We will consider mentioning file downloading in handler examples (thanks for pointing out that this is unclear how to achieve it). I think a dedicated section for file downloading is too much, as it is not a common use case in plugins using JCEF.

timKraeuter:
Thanks for considering that. I have two other remarks which might be interesting:

- I think the page deserves a better name. JCEF is cryptic and might mean little to plugin developers looking for a way to integrate web applications into the IDE. Maybe something like **"Web View using JCEF"** or **"Embedded Browser (JCEF)"** would be more precise.
- An example of how to make a JCEF panel freely resizable or fill the whole tab in the IDEA would also be greatly appreciated. I plugin developers opt for a web-based UI. In my opinion, this is what they are looking for.

I am looking forward to seeing the changes in the internal review soon!

PR #1180, opened 2023-12-14
PlantUML diagrams inlined

karollewandowski:
Styles must be inlines until global style support is implemented. I minimized the styles only to those required by a given diagram.

The getting service diagram has been reworked:
- it is more narrow and doesn't need a thumbnail
- texts are cleaner (also grammar issues were fixed)


YannCebron (commented):
remove `buildUML/README.md`

YannCebron (approved):

PR #1185, opened 2024-01-05
Document `RestLanguage` moved to another package as part of V2 migration

throwable-one:
As part of migration to v2 modules we moved this class to another package.

See `4da3e7da0dcd86d454972b2ef68d2581c7fe5cdb`


YannCebron:
6437fa3fb84d7e674adad0d111945343e3f17e61

PR #1186, opened 2024-01-09
Document `com.jetbrains.commandInterface` moved to another package as part of V2 migration

throwable-one:
As part of migration to v2 modules we moved this class to another package.

See `18dcc57922878c8ad1fe0bb6be85097d4d81cc34`


YannCebron:
wait for fix

PR #1187, opened 2024-01-09
`com.jetbrains.jinja2` package renamed as part of v2

throwable-one:


YannCebron:
Pattern not supported, fixed in 616f1ee805adfbf8f6c468139a5041a187344501

PR #1248, opened 2024-02-24
Fixed comment string.

azhidkov:
In the source code 
`// ** Tutorial step 19. - Add a quick fix for the str`
but in documentation Tutorial step 18.3. Fixed it.


YannCebron:
Thanks for spotting!

PR #1251, opened 2024-03-07
Update api_changes_list_2024.md

dmitry-avdeev:
package removed


YannCebron:
This is not a valid pattern. What classes are affected here in addition to the one already listed?

dmitry-avdeev:
I've just moved a class, changing its package. "class moved" is already documented, the whole change would be

### Markdown Plugin 2024.1
`org.intellij.plugins.markdown.ui.split` package removed
`org.intellij.plugins.markdown.ui.split.SplitTextEditorProvider` class removed
: `com.intellij.openapi.fileEditor.TextEditorWithPreviewProvider` should be used instead

dmitry-avdeev:
Without "package removed" I still get

Package 'org.intellij.plugins.markdown.ui.split' is not found

YannCebron:
I see. I'll add correct entry and mute test.

PR #1293, opened 2024-04-05
Fix incorrect link in EditorBasedStatusBarPopup

ruyut:
Fix incorrect link in EditorBasedStatusBarPopup


karollewandowski:
Thank you for the fix 🙂

PR #1298, opened 2024-04-11
Add note about kotlin update call

gauntface:
I'm new to IntelliJ dev and Kotlin so please feel free to close this if it's not helpful, but this would have save me an hour or two trying to understand why `updateWidget()` calls were not updating the status bar.


karollewandowski:
Hi Matt. I'm not sure I understand why it wasn't clear to you.
If it was the Kotlin's [passing trailing lambdas](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas) syntax, then I think we shouldn't include it on the page (we would have to do this in hundreds of places).

Anyway, thank you for the PR.

PR #1301, opened 2024-04-17
Fixes for the IJP Gradle Plugin 2.0

halirutan:
Some things to discuss so far:

- [ ] `showPaidPluginWarning` description does not explain why this is needed. It just explains what it does, which everyone can guess from the name already.
- [x] Sometimes, we have task descriptions starting as real sentences like "This is a task used to..."; other times, we just write "Prints the list of...". What form should we use consistently?
- [ ] `verifyPluginConfiguration` Is it correct that this task validates that "The currently selected Java Runtime is not JetBrains Runtime (JBR)"?
- [x] We are inconsistent and use paragraphs basically as bullet points. [E.g. these are not full sentences](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-tasks.html#verifyPlugin-offline ) but they could be one paragraph which I would much prefer. Can we make this consistent?

### DIFF
--- a/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
+++ b/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
@@ -857,7 +857,7 @@ Type
 
 </tldr>
 
-This is a task used to prepare an immutable `test` task and provides all necessary dependencies and configurations for a proper testing configuration.
+This is a task used to prepare an immutable `test` task and provide all necessary dependencies and configurations for a proper testing configuration.
 
 
 ## `printBundledPlugins`
--- a/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
+++ b/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
@@ -703,7 +703,7 @@ See also:
 {#patchPluginXml-untilBuild}
 
 The highest IDE version compatible with the plugin.
-The `until-build` attribute can be unset by setting `provider { null }` as a value, and not that only passing `null` will make Gradle use the default value instead.
+The `until-build` attribute can be unset by setting `provider { null }` as a value, and note that only passing `null` will make Gradle use the default value instead.
 However, if `until-build` is undefined, compatibility with all the IDEs since the version specified by the `since-build` is assumed, which can cause incompatibility errors in future builds.
 
 The provided value will be assigned to the `<idea-version until-build="..."/>` element attribute.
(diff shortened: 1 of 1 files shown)


karollewandowski:
My opinion:

> `showPaidPluginWarning` description does not explain why this is needed. It just explains what it does, which everyone can guess from the name already.

+1; it would be nice to explain the purpose.

> Sometimes, we have task descriptions starting as real sentences like "This is a task used to..."; other times, we just write "Prints the list of...". What form should we use consistently?

I think we should use more concise form, so "Prints the list..." instead of "This task prints the list...". This style is used in, e.g., method Javadocs, which I think are analogical to task descriptions on our pages.

> We are inconsistent and use paragraphs basically as bullet points. [E.g. these are not full sentences](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-tasks.html#verifyPlugin-offline) but they could be one paragraph which I would much prefer. Can we make this consistent?

I don't understand. It looks good to me, but I don't get the lack of the period in the second one.
Could you please provide an example sentence?

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
```diff
@@ -328,11 +331,10 @@ Type
 
 </tldr>
 
-Task dedicated to executing the code instrumentation using the Ant tasks provided with the currently used IntelliJ Platform dependency.
+Task dedicated to executing the code instrumentation using the Ant tasks provided by the used IntelliJ Platform dependency.
```
karollewandowski, line 334:
```suggestion
Executes the code instrumentation using the Ant tasks provided by the used IntelliJ Platform dependency.
```

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_tasks.md
```diff
@@ -559,11 +555,10 @@ Default value
 ### `pluginVersion`
 {#patchPluginXml-pluginVersion}
 
-The plugin version is displayed in the <control>Plugins</control> settings dialog and on the JetBrains Marketplace plugin page.
+The plugin version is displayed in the <control>Plugins</control> settings dialog and on the JetBrains Marketplace plugin page,
+and plugins uploaded to the JetBrains Marketplace must follow [semantic versioning](https://plugins.jetbrains.com/docs/marketplace/semver.htm).
```
karollewandowski, line 559: I would keep two simpler sentences (a general rule for copywriting is to try to make sentences short). Removing a separate paragraph is fine to me.

karollewandowski (approved):
Looks good to me (two minors).

YannCebron:
Please note that some texts need to be synced with kdoc in the plugin itself.

hsz:
Thanks!

PR #1357, opened 2024-07-29
IntelliJ Platform Gradle Plugin (2.X) first page only

Jonatha1983:
Hi 

This PR only contains changes on the IntelliJ Platform Gradle Plugin (2.X) page.

**Those are my general suggestions and opinions (feel free to disagree 😄 ):**
1. This page at the end will be for all users who want to write a plugin for JB, ide, and I think that for new users, the migration sections and the various linkages to version 1.X can be overwhelming.
It can start with a note and a link to users who are migrating from previous version pages, but it should not mention migration and 1.X more than that.    
2. The fact that this Gradle **plugin** is for creating an IDE **plugin** causes some confusion when using the plugin noun. So, I think it is always better to be very clear about the meaning, even if it is clear for most of the people from the context.
3. I find that the structure of this specific page is a bit confusing because it has two main topics: Usage and Configuration. When I went over it, I thought that some of the usage parts should be related to the configuration, while some of the configuration parts should be related to the usage. In addition,, the section of the sub-Gradle plugins should be in a separate part dedicated to more internal and in-depth work.

### DIFF
--- a/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
+++ b/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
@@ -14,7 +14,7 @@
 
 </tldr>
 
-The _IntelliJ Platform Gradle Plugin 2.x_ is a plugin for the Gradle build system to help configure environments for building, testing, verifying, and publishing plugins for IntelliJ-based IDEs.
+The _IntelliJ Platform Gradle Plugin 2.x_ is a [Gradle](https://docs.gradle.org/current/userguide/userguide.html) plugin for building, testing, verifying,configure environments, and publishing IntelliJ-based IDEs plugins.
 It is a successor of _[](tools_gradle_intellij_plugin.md)_.
 
 <snippet id="faq">
@@ -27,7 +27,8 @@ It is a successor of _[](tools_gradle_intellij_plugin.md)_.
 
 ## Requirements
 
-IntelliJ Platform Gradle Plugin 2.x requires the following minimal versions:
+IntelliJ Platform Gradle Plugin 2.x requires the following **minimal versions**:
+
 - IntelliJ Platform: **2022.3**
 - Gradle: **8.2**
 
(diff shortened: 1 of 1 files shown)


topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
```diff
@@ -27,7 +27,8 @@ It is a successor of _[](tools_gradle_intellij_plugin.md)_.
 
 ## Requirements
 
-IntelliJ Platform Gradle Plugin 2.x requires the following minimal versions:
+IntelliJ Platform Gradle Plugin 2.x requires the following **minimal versions**:
```
Jonatha1983, line 30: I made the minimal in bold, but I think it will be better to write the '+' or "and above" next to each requirement

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
```diff
@@ -42,19 +43,28 @@ IntelliJ Platform Gradle Plugin 2.x requires the following minimal versions:
...
 
-To apply the IntelliJ Platform Gradle Plugin to a project, add the following entry to the `plugins` block in the <path>build.gradle.kts</path> file:
-
-```kotlin
-plugins {
-  id("org.jetbrains.intellij.platform") version "%intellij-platform-gradle-plugin-version%"
-}
-```
-
+To apply the IntelliJ Platform Gradle Plugin to a project, you need to add the following:
```
Jonatha1983, line 46: I'm not sure, but I think the user needs to add the gradle portal repository to use the plugin. If I am wrong, this can be reverted.

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
```diff
@@ -42,19 +43,28 @@ IntelliJ Platform Gradle Plugin 2.x requires the following minimal versions:
...
+   </code-block>
+
+<tip>
 If migrating from the Gradle IntelliJ Plugin 1.x, replace the old `org.jetbrains.intellij` identifier to `org.jetbrains.intellij.platform` and apply its latest `%intellij-platform-gradle-plugin-version%` version.
+</tip>
 
 ### Snapshot Release
 
-To use the latest snapshot versions, add the following to the <path>settings.gradle.kts</path> file:
+To use the latest snapshot versions of this plugin, add the following to the <path>settings.gradle.kts</path> file:
```
Jonatha1983, line 67: Between all the definitions and usages of IJ target products and builds with snapshots and releases, this sentence should emphasize that we are talking here about the plugin version

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
```diff
@@ -72,24 +82,28 @@ pluginManagement {
...
 
-
 ### Attaching Sources
 
 To attach IntelliJ Platform sources in the IDE, the <control>Download sources</control> setting has to be enabled in IDE versions 2023.2 and later.
 This option respects the [](tools_intellij_platform_gradle_plugin_gradle_properties.md#downloadSources) property, which is enabled by default.
 
 <tabs>
 
+<tab title="Earlier versions">
```
Jonatha1983, line 101: I thought those should be in chronological order since it started with the 2023.3+ tab and then a tab of 2023.2, and jumping to the earlier version is a bit confusing.

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
```diff
@@ -104,17 +118,11 @@ Then invoke the <control>Reload All Gradle Projects</control> action from the <c
...
-<tab title="Earlier versions">
-
-No additional IDE settings are required.
-
-</tab>
-
 </tabs>
 
-The attaching sources is handed by the Plugin DevKit plugin thus it's recommended to use always the latest available IDE release.
+The attaching sources operation in the IDE is handled by the Plugin DevKit plugin thus it's always recommended to use the latest available IDE release.
```
Jonatha1983, line 123: I found this a bit unclear. 

The DevKit plugin handled the attaching sources so it is recommended to use the latest plugin version available or the latest IDE release?

topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin.md
YannCebron: It _is_ in chronological order. Latest version on the left, oldest on the right; this is also consistent with all other such tabs in docs. We assume that plugin developer always upgrade to new IDEs ASAP.

YannCebron:
Thanks a lot for your PR and sorry for the long delay. I've incorporated the (still relevant) changes to the current version manually.

PR #1367, opened 2024-08-23
Describe required JSON plugin dependency in API changes doc

NikitaKatkov:
JSON plugin was recently extracted from the platform, so external plugins have to add an explicit dependency on it to be able to reuse its classes.

Some of the reports produced by our API watcher might be incorrect because the plugin itself has not been yet published to the marketplace. Anyways, some of the external plugins do lack the necessary dependency, so the fix is described in the sdk docs


karollewandowski:
Thanks.

PR #1368, opened 2024-09-02
Update usage of pluginName in tools_intellij_platform_gradle_plugin_recipes.md

AleksandrSl:
Maybe I understood I did the update to 2.0 incorrectly, but `pluginName` doesn't work for me, but `intellijPlatform.pluginConfiguration.name` (described [here](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-migration.html#intellijpluginname)) works fine. 
I didn't test groovy version, since I only use kts, but presume it should the same

As a sidenote for the plugin template, is it on purpose that only the [version is assigned explicitly](https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/build.gradle.kts#L56), but the name isn't. I had to assign it as well for the property to work

### DIFF
--- a/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_recipes.md
+++ b/topics/appendix/tools/intellij_platform_gradle_plugin/tools_intellij_platform_gradle_plugin_recipes.md
@@ -108,7 +108,7 @@ Additional files can be bundled by adding them to the plugin directory when prep
 tasks {
   prepareSandbox {
     from(layout.projectDirectory.dir("extraFiles")) {
-      into(pluginName.map { "$it/extra" })
+      into(intellijPlatform.pluginConfiguration.name.map { "$it/extra" })
     }
   }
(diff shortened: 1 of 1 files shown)


hsz:
Thank you for reporting.

This is actually working as expected as the `intellijPlatform.pluginConfiguration.name` property is supposed to be a **name of the plugin** – which is passed to the `plugin.xml` file, and eventually visible in JetBrains Marketplace.

To configure the **archive name**, you are supposed to use the `intellijPlatform.projectName`, which takes the `project.name` value as a default value and passes it further to the `BuildPluginTask.archiveName`.

AleksandrSl:
@hsz thanks for checking this. However in the example `pluginName` is used, not `projectName`

Exactly as in the recipe
Screenshot 2024-09-26 125615

image


Your suggestion to use `projectName` works well indeed
Screenshot 2024-09-26 125536

Sorry if I miss some other context

PR #1417, opened 2025-01-21
code_intentions.md: edit considering Intentions Actions Preview feature

vladimir-petrenko:
purpose:
- make it clearer that "before/after examples" and "intention action preview" are different things.
- simplify navigation to "Intention Action Preview" topic


YannCebron:
Thanks!

PR #1431, opened 2025-03-09
Added extension point information for customizing action-based toolbars

slideclimb:
Added extension point information to the [Customization](https://plugins.jetbrains.com/docs/intellij/toolbar.html#customization) section in the toolbar guidelines.

Another option is to add this information to [Building a Toolbar/Popup Menu from Actions](https://plugins.jetbrains.com/docs/intellij/basic-action-system.html#buildingToolbarPopupMenu) instead, as that is more about implementation details. Let me know which location you prefer and I'll move it if needed.


karollewandowski (changes requested):
Hi,
Thank you for the PR.

I think it should be moved to the Actions page. We link the "Building a Toolbar/Popup Menu from Actions" section at the top of the Toolbar page focused on UI.

Please also rephrase the sentence to something like:
> ```
> To add an action group to the list of customizable actions in <ui-path>Settings | Appearance & Behavior | Menus and Toolbars</ui-path>, implement the [`CustomizableActionGroupProvider`](%gh-ic%/platform/platform-impl/src/com/intellij/ide/ui/customization/CustomizableActionGroupProvider.java) extension point registered in `com.intellij.customizableActionGroupProvider`, and ensure that the [action group](plugin_configuration_file.md#idea-plugin__actions__group) defines the `text` attribute or is [localized](basic_action_system.md#localizing-actions-and-groups).
> ```

topics/ui/controls/toolbar.md
```diff
@@ -195,4 +195,3 @@ In places other than tool windows, it is fine to always show the border.
 ## Customization
 
 Provide the ability to customize toolbars in the main window. To do this, add the toolbar to the list on the <ui-path>Settings | Appearance & Behavior | Menus and Toolbars</ui-path> page.
-
```
karollewandowski, line 198: Please revert this change.

karollewandowski (changes requested):

karollewandowski:
Thanks again 🙂

PR #1433, opened 2025-03-11
Update api_changes_list_2025.md

HackerMadCat:
The ExternalSystemProcessingManager class was extended from ExternalSystemTaskNotificationListener and DIsposable. Due to this, their functions were exposed in public API, when it is implementation detail of the ExternalSystemProcessingManager service.


YannCebron:
the proper pattern is `<class name> class now interface` see comments on top of the page

PR #1435, opened 2025-03-11
Document changes in the Analysis API in 2025.1

yanex:
In 2025.1, there were changes in the Analysis API related to previously deprecated and experimental endpoints.


YannCebron:
Deprecations and experimental API changes do not need to be documented

PR #1451, opened 2025-03-31
Update ide.tooltip.initialReshowDelay value

hamen:
Looking at the current source code, the value is 500.

For reference https://github.com/JetBrains/intellij-community/blob/master/platform/util/resources/misc/registry.properties#L328


karollewandowski:
@MariannaKononenko, is it correct? Can we merge it?

PR #1452, opened 2025-04-02
Fix `noSearchableOptionsWarning` example

strangelookingnerd:
I noticed the example references `buildSearchableOptions` instead of `noSearchableOptionsWarning`.


strangelookingnerd:
@hsz Kindly requesting a review.

PR #1453, opened 2025-04-04
empty_state_marianna

MariannaKononenko:


topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
-    <tr>
-        <td><img src="libraries-before.png" alt="" width="241" /> Avoid &quot;Nothing to show&quot; as it is not informative.</td>
-        <td><img src="libraries-after.png" alt="" width="241" /></td>
-    </tr>
+Empty states of tool windows, lists, trees, tables, master-detail layouts, or other container components should be informative for the users.
+Fill in the empty states with the following:
+- [A reason why the container is empty](#reason-why-empty)
+- [Actions to fill in the area](#actions-to-fill-the-area)
+- [A link to the corresponding help topic](#help-topic)
+
```
OlyaB, line 18: I really liked how the article is restructured now, thank you!

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+### Explain the current state
+
+The default pattern is `No [entity] added.` If `added`, `created`, `configured`, or other words act as synonyms in a particular case, use `added` for consistency.
+
+<table style="none" border="false" column-width="fixed">
+  <tr>
+    <td>
+      <format color="369650" style="bold">Correct</format>
+      <img src="empty_state_clear_reason.png" alt="Correct text in the empty state" />
```
OlyaB: Should these pics be about the library not the SQL dialects? Otherwise, we have the same illustration in nearby sections

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+      <format color="E55765" style="bold">Incorrect</format>
+      <img src="empty_state_unclear_reason.png" alt="The reason is not clear"/></td>
+  </tr>
+</table>
 
-If an action cannot be tied to a link, explain what to do.
+## Actions to fill the area
 
-![](todo-tw.png){width=574}
+Required part. An action makes it easier to understand what to start with
```
OlyaB: "Required part" — possibly write as "The only required part" to make it more clear that other parts are optional?

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+    </td>
+    <td>
+      <format color="E55765" style="bold">Incorrect</format>
+      <img src="empty_state_menu_incorrect.png" alt="Menu under the action"/>
+    </td>
+  </tr>
+</table>
 
-**Tool windows:** provide a link to a help topic that explains the functionality. Add the question mark icon in the beginning.
+### Explain what to do when there is no action`
```
OlyaB, line 105: Extra backtick symbol at the end of the header

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+
+<p>Do not introduce new entities that are not used in the current context.
+Instead, use the terms from the current UI so that the user understands
+what exactly is missing:</p>
+
+<table style="none" border="false" column-width="fixed">
+  <tr>
+    <td>
+      <format color="369650" style="bold">Correct</format>
+      <img src="empty_state_clear_reason.png" alt="The reason is clear"/>
```
OlyaB, line 50: The half-width pics look a bit blurry in the article, possibly need to explicitly specify the width 378px

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+<table style="none" border="false" column-width="fixed">
+  <tr>
+    <td width="378">
+      <img src="empty_state_link_to_help.png" alt="Link to a help topic"/>
+    </td>
+    <td>
+    <control>Link to a help topic</control>
+    <p>In tool windows and <a anchor="empty-state-of-the-master-detail-layout">master-detail layouts</a>,
+provide a link to a help topic that explains the functionality.
+Add the question mark icon in the beginning.</p>
```
OlyaB, line 136: Possibly add a link to the question mark icon so it's obvious which one https://intellij-icons.jetbrains.design/#AllIcons-expui-general-questionMark

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+  </tr>
+  <tr>
+    <td width="378">
+      <img src="empty_state_inline_help.png" alt="Inline hint under the table"/>
+    </td>
+    <td>
+      <control>Context help in inline hints</control>
+      <p>In tables, trees, and lists, provide instructions according to the <a href="context_help.md" anchor="inline-help-text">inline text help</a> rules.
+Smaller UI areas rarely require an in-depth explanation compared to complex tool windows.
+A short help text should be enough, and it does not require switching contexts.</p>
```
OlyaB: Didn't understand why we have these two last sentences here until I carefully read again the previous part — possibly specify more explicitly that the question mark link is only for tool windows and master-details, and not for any other place.

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
-![](todo-tw.png){width=574}
+Required part. An action makes it easier to understand what to start with
+instead of searching for the appropriate icon on the toolbar.
+It can also educate about the shortcut.
 
-Hide the area's toolbar if it does not have the same action as in the empty state. Usually, all other toolbar actions are not relevant in this case.
+### Use one or two actions
 
-<format color="Red" style="bold">Incorrect</format>
+Three or more actions would make it harder to choose what to start from.
```
OlyaB, line 66: We have a couple of cases where we list 3 actions, like in the Run tool window. And I think it's an ok case to be mentioned. Possibly the guideline could be "In rare cases, when you need to list all possible starting points, use the list layout."

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
-<format color="Red" style="bold">Incorrect</format>
+- If the master area is empty, display the explanation, action, and link to the help article.
+- If the details area is empty, no explanations are needed. In this case, show the actions to fill in the area.
+  The default pattern for the action is "_Select_ [entity] _to configure_".
 
-![](todo-filters-incorrect.png){width=456 style=block}
-*Do not use a help topic link in a table and UI areas other than tool windows.*
+<img src="empty_state_master_detail_layout.png" alt="Empty master-detail layout" width="706"/>
 
-<format color="Green" style="bold">Correct</format>
```
OlyaB, line 161: There was one more pic in the old article with the master part having a list https://plugins.jetbrains.com/docs/intellij/empty-state.html#master-detail-layout.

Possibly, add the 3rd bullet point with a guideline something like "When there's content in the master part, provide a specific empty state in the detail area"?

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+    <td>
+      <format color="369650" style="bold">Correct</format>
+      <img src="empty_state_verbs_correct.png" alt="Correct usage of verbs in the empty state"/>
+    </td>
+    <td>
+      <format color="E55765" style="bold">Incorrect</format>
+      <img src="empty_state_verbs_incorrect.png" alt="Incorrect usage of verbs in the empty state"/>
+    </td>
+  </tr>
+</table>
```
OlyaB, line 216: Possibly return info about non-breaking spaces? https://plugins.jetbrains.com/docs/intellij/empty-state.html#sizes-and-placement

It's a small detail but makes the typography a bit more polished

OlyaB (changes requested):
Great update! Also left some comments for pics cosmetics in Figma

topics/ui/principles/empty_state.md
eldar-jetbrains: Maybe put it above all other chapters that are not required if this one is the most important?

eldar-jetbrains: Maybe even put it in a note component to emphasise the importance? Right now it's easy to miss

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
 
-### Writing guidelines
+## Writing guidelines
 
-See [Punctuation](punctuation.md) and [Capitalization](capitalization.md).
+### Punctuation
 
-Make the reason text short and descriptive. See [Writing short and clear](writing_short.md).
+Separate sentences in the empty state with periods, but do not put a period after the action link.
+Use the ellipses at the end of the action link if this link opens a dialog or a popup which requires some input from the user.
```
eldar-jetbrains, line 168: Maybe add a link to the Link guideline? There all rules are covered about how to write links

topics/ui/principles/empty_state.md
```diff
@@ -10,144 +10,261 @@
...
+      <img src="empty_state_capitalization_incorrect.png" alt="Incorrect capitalization in the empty state"/>
+    </td>
+  </tr>
+</table>
 
-Avoid saying _add new_. Just use _add_ because all that is added is new in the context of an empty UI area.
+### Avoid excessive verbs
 
-## Master-detail layout
+- In actions, avoid words that describe physical actions like _press_ or _click_ — they are obvious from how the link works.
```
eldar-jetbrains: If I understood correctly the https://plugins.jetbrains.com/docs/intellij/how-to-write-guidelines.html#word-level-recommendations, we should use <code> to format quotes instead of using italic. In this case all the quotes should be formatted differently

topics/ui/principles/empty_state.md
MariannaKononenko: How do we decide which part is required and which is optional? I'd say both reason and action should be required, and only help is optional. Maybe omit this phrase about the required part at all? Just mention that the help part can be skipped. As for changing the order, I would not do it because the current order reflects how these parts are placed in the ui, one by one

topics/ui/principles/empty_state.md
MariannaKononenko: Maybe suggest the layout that was recently implemented for the Find tool window?

topics/ui/principles/empty_state.md
MariannaKononenko: here is not clear: should we give an action link in the details area in this case (which is actually duplicating the action in the master area) or just add the gray text?

PR #1459, opened 2025-04-29
Update tools_intellij_platform_gradle_plugin_repositories_extension.md - `CustomPluginRepositoryType` -> `CustomPluginRepositoryListingType`

ChrisCarini:


hsz:
Both versions are valid. The IntelliJ Platform Gradle Plugin introduces an alias:

https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/6c7de8d48aeecc390ea98f2c9e3b103e3016f518/src/main/kotlin/org/jetbrains/intellij/platform/gradle/CustomPluginRepositoryType.kt#L7

PR #1505, opened 2025-07-08
IJPL-175712 Accessibility UI guidelines

dmitrii-drobotov:


topics/ui/principles/accessibility.md
```diff
@@ -0,0 +1,340 @@
...
+
+## Keyboard accessibility
+
+All functionality must be operable by using only the keyboard.
+People who have trouble using a mouse, including those who are blind, have low vision, or have motor disabilities, rely on keyboard navigation to effectively use the interface.
+
+### Focus basics
+
+The basic way to interact with the UI by using the keyboard is to switch focus between components with <shortcut>Tab</shortcut> and <shortcut>Shift+Tab</shortcut>.
+For some components, arrow keys can also be used to navigate between items.
```
MariannaKononenko: maybe give examples of the components that can be only reached by clicking arrows (list, drop-down menus, etc)?

topics/reference_guide/accessibility.md
```diff
@@ -0,0 +1,340 @@
...
+## Keyboard accessibility
+
+All functionality must be operable by using only the keyboard.
+People who have trouble using a mouse, including those who are blind, have low vision, or have motor disabilities, rely on keyboard navigation to effectively use the interface.
+
+### Focus basics
+
+The basic way to interact with the UI by using the keyboard is to switch focus between components with <shortcut>Tab</shortcut> and <shortcut>Shift+Tab</shortcut>.
+For some components, arrow keys can also be used to navigate between items.
+Once a component is focused, it can be activated by pressing <shortcut>Space</shortcut>.
```
MariannaKononenko, line 20: also add info on how focused components should look like (e.g. get highlighted with the blue stroke)

topics/reference_guide/accessibility.md
```diff
@@ -0,0 +1,340 @@
...
+However, in some cases, they can be focusable to let screen reader users access important information.
+Use [`ScreenReader.isActive()`](%gh-ic%/platform/util/ui/src/com/intellij/util/ui/accessibility/ScreenReader.java) to adjust the behavior only for screen reader users.
+
+### Keyboard traps
+
+Make the UI traversable in a cycle without trapping the keyboard focus.
+A [keyboard focus trap](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) occurs when a user cannot navigate away from a component by using the standard keyboard navigation.
+This prevents them from accessing other parts of the interface.
+
+Ensure that a user can always navigate away from any focusable component by using <shortcut>Tab</shortcut>, <shortcut>Shift+Tab</shortcut>, or <shortcut>Escape</shortcut>.
```
MariannaKononenko, line 35: an example could be helpful here, e.g. how to navigate from the project tool window to the editor (?)

topics/ui/principles/accessibility.md
```diff
@@ -0,0 +1,340 @@
...
+This integration is enabled by properly defined accessible metadata.
+
+Follow these guidelines to ensure that users of assistive technology can fully interact with the UI.
+
+### Accessible properties
+
+Each UI component has an associated [`javax.accessibility.AccessibleContext`](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/javax/accessibility/AccessibleContext.html) object that defines properties for assistive technologies.
+The accessible context can also implement [`Accessible*`](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/javax/accessibility/package-summary.html#class-summary) interfaces to provide additional metadata or ways for assistive technologies to interact with the component.
+
+> For common UI components, basic accessibility support is often already implemented.
```
MariannaKononenko: common UI components  - add a link to Components.topic

topics/reference_guide/accessibility.md
```diff
@@ -0,0 +1,340 @@
+<!-- Copyright 2000-2025 JetBrains s.r.o. and contributors. Use of this source code is governed by the Apache 2.0 license. -->
+
+# Accessibility
+
```
MariannaKononenko, line 4: Since the article is mainly focused on iplementation details, maybe it  makes sense to move it for example to the Base Platform section?

dmitrii-drobotov: Good idea. I think I can extract the Assistive technology support section into an article here: https://plugins.jetbrains.com/docs/intellij/user-interface-components.html. And we can keep things about keyboard accessibility in UI guidelines and also later add the information about colors here. Both articles will have links to each other. What do you think?

topics/reference_guide/accessibility.md
dmitrii-drobotov: I can now add some basic info (1-2 sentences) that focused components need to have a focus indicator. And then later you can expand it with an illustration and add more info if needed. I wrote about this part here: [UX-3531](https://youtrack.jetbrains.com/issue/UX-3531). Does that sound good?

topics/reference_guide/accessibility.md
OlyaB: > keyboard accessibility in UI guidelines

I don't think it's worth separating this article into two ones. I'm also not sure it should be on the level of UI. Accessibility is an important concept that includes a variety of principles. I'd suggest placing it right under "Base Platform". To me, it's similar to "Internationalization" which is also about making the UI accessible.

When we have something about colors, we can then see where to place it. I think we should provide a default palette / color schemes that already incorporate the color accessibility principles, so that plugin developers just need to use the correct color tokens.

OlyaB (approved):
Answered to a comment

topics/reference_guide/accessibility.md
dmitrii-drobotov: Sounds good, I moved the article to the Base Platform and also removed the "Accessibility" link from the UI Guidelines  - Principles.

PR #1899, opened 2026-03-26
Update light_and_heavy_tests.md - remove reference to deprecated `LightPlatformTestCase`

ChrisCarini:
Ideally, the docs wouldn't guide a user to use a class that's deprecated - I recommend removing this reference so that developers are guided to use something 'current' and not deprecated.


karollewandowski:
Thank you for the PR, but it has to wait until we decide whether this class actually should be deprecated, and until we properly document new alternatives in mentioned JUnit 5.

PR #2021, opened 2026-06-10
Add a code sample for multi-module plugin and the Gradle build setup

novotnyr:
Add a code sample for multi-module plugin with Gradle setup.

See README.md for technical info and layout.

This is a tie-in to [Gradle Setup Powering Multi-module IntelliJ Plugins](https://youtu.be/7fEdrktCYDo) video.

PR #2022, opened 2026-06-10
Fix missing plugin metadata in OAuth 2.0 Sample

novotnyr:
Plugin version and description is mandatory, so add them. Align name and group with other plugins.

This fixes failing tests when Plugin Verifier fails on descriptor validation.


hsz (approved):

hsz:
Thank you

novotnyr:
This also unblocks failing tests in #2021.

karollewandowski (approved):

PR #2044, opened 2026-06-24
AT-3202 Updated Integration Test Documentation

Artem-Grigorash:
Made amendments to replace outdated and non-functional code and added some details for greater clarity.

PR #2101, opened 2026-07-12
Refactor plugin installation to use newer installPluginFromDir

AleksandrSl:
First example already uses the newer installPluginFromDir method. Examples better be consistent I think.

If the bridge to the older method is needed it could be written as a note, like zinstallPluginFromFolder is the deprecated way to do this, use installPluginFromDir nowz, but I think it's too much given IDE has the warning already

### DIFF
--- a/topics/basics/testing_plugins/integration_tests/integration_tests_intro.md
+++ b/topics/basics/testing_plugins/integration_tests/integration_tests_intro.md
@@ -133,7 +133,7 @@ The test case uses IntelliJ IDEA version 2024.3, and starts the IDE without any 
 ```kotlin
 .apply {
   val pathToPlugin = System.getProperty("path.to.build.plugin")
-  PluginConfigurator(this).installPluginFromFolder(File(pathToPlugin))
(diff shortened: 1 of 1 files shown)


topics/basics/testing_plugins/integration_tests/integration_tests_intro.md
AleksandrSl, line 103: Actually, install from dir doesn't work in my setup at all, thought it's the most basic one, because `pathToPlugin ` is a path to the zip, and `installPluginFromDir` expects a directory containing the zip/or plugin files. So it creates a directory that is named the same way as a zip file and then fails.

What works for me is 
`PluginConfigurator(this).installPluginFromPath(Path.of(pathToPlugin))`
I wonder if it's only my problem

karollewandowski:
Hi @AleksandrSl , it seems to be covered in https://github.com/JetBrains/intellij-sdk-docs/pull/2044/changes.

AleksandrSl:
Yep, it doesn't clarify on what is supposed to be copied and why in my case the target of copy is zip file, but could be my setup problem.

Thanks for the updates!

PR #2131, opened 2026-07-24
Improve and expand the Persisting Sensitive Data section

novotnyr:
Improve and expand on the Persisting Sensitive Data section.

Provide common scenarios for

- passwords
- API keys (no username)
- private keys (binary content)

Provide both Kotlin and Java code in every example.

Additionally, provide a Kotlin-based `@Service` implementations with `suspend fun` to indicate proper thread usage.

### DIFF
--- a/topics/basics/persisting_sensitive_data.md
+++ b/topics/basics/persisting_sensitive_data.md
@@ -26,7 +26,8 @@ Pass a subsystem name that identifies the plugin or its area of functionality, a
 ```kotlin
 import com.intellij.credentialStore.generateServiceName
 
-val serviceName = generateServiceName("My Password Storage", "john.doe")
+val serviceName =
+    generateServiceName("My Password Storage", "john.doe")
 ```
 </tab>
 <tab title="Java" group-key="java">
@@ -59,10 +60,10 @@ To store a username, account name or other identity, set it along with the servi
 ```kotlin
 import com.intellij.credentialStore.generateServiceName
 
-private const val SERVICE_NAME = "My Password Storage"
 //...
+
 private fun credentialAttributesOf(username: String): CredentialAttributes {
-  val serviceName = generateServiceName(SERVICE_NAME, username)
+  val serviceName = generateServiceName("My Password Storage", username)
   return CredentialAttributes(serviceName, username)
 }
 ```
@@ -73,12 +74,11 @@ private fun credentialAttributesOf(username: String): CredentialAttributes {
 ```java
 import com.intellij.credentialStore.CredentialAttributesKt;
 
-private static final String SERVICE_NAME = "My Password Storage";
-
 // ...
 
 private static CredentialAttributes credentialAttributesOf(String username) {
-  String serviceName = CredentialAttributesKt.generateServiceName(SERVICE_NAME, username);
+  String serviceName = CredentialAttributesKt
+      .generateServiceName("My Password Storage", username);
   return new CredentialAttributes(serviceName, username);
 }
 ```
@@ -102,9 +102,11 @@ To store a `String`-based password, use the `PasswordSafe` instance:
 ```kotlin
 @Service
 class PasswordService {
-  suspend fun save(username: String, password: String) = withContext(Dispatchers.IO) {
-    val credentialAttributes: CredentialAttributes = credentialAttributesOf(username)
-    PasswordSafe.instance.setPassword(credentialAttributes, password)
+  suspend fun save(username: String, password: String) {
+    withContext(Dispatchers.IO) {
+      PasswordSafe.instance
+          .setPassword(credentialAttributesOf(username), password)
+    }
   }
 
   private fun credentialAttributesOf(username: String): CredentialAttributes {
@@ -120,8 +122,8 @@ class PasswordService {
 public final class PasswordService {
   @RequiresBackgroundThread
   public void save(String username, String password) {
-    CredentialAttributes credentialAttributes = credentialAttributesOf(username);
-    PasswordSafe.getInstance().setPassword(credentialAttributes, password);
+    PasswordSafe.getInstance()
+        .setPassword(credentialAttributesOf(username), password);
   }
 
   private CredentialAttributes credentialAttributesOf(String username) {
@@ -151,8 +153,7 @@ To retrieve a stored password:
 
 ```kotlin
 suspend fun find(username: String): String? = withContext(Dispatchers.IO) {
-  val credentialAttributes: CredentialAttributes = credentialAttributesOf(username)
-  PasswordSafe.instance.getPassword(credentialAttributes)
+  PasswordSafe.instance.getPassword(credentialAttributesOf(username))
 }
 ```
 </tab>
@@ -161,8 +162,8 @@ suspend fun find(username: String): String? = withContext(Dispatchers.IO) {
 ```java
 @RequiresBackgroundThread
 public String find(String username) {
-  CredentialAttributes credentialAttributes = credentialAttributesOf(username);
-  return PasswordSafe.getInstance().getPassword(credentialAttributes);
+  return PasswordSafe.getInstance()
+      .getPassword(credentialAttributesOf(username));
 }
 ```
 </tab>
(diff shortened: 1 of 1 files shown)


topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
-  return CredentialAttributes(generateServiceName("MySystem", key))
+import com.intellij.credentialStore.generateServiceName
+
+val serviceName = generateServiceName("Password Storage Showcase", "john.doe")
+```
+</tab>
+<tab title="Java" group-key="java">
+
+```java
+String subsystem = "Password Storage Showcase";
```
karollewandowski: Please inline, like in the Kotlin snippet. If it doesn't fit the chars, break lines.

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
 
 </tab>
 <tab title="Java" group-key="java">
 
 ```java
-private CredentialAttributes createCredentialAttributes(String key) {
-  return new CredentialAttributes(
-    CredentialAttributesKt.generateServiceName("MySystem", key)
-  );
+private static final String SERVICE_NAME = "Password Storage Showcase";
```
karollewandowski: Please add import here as well for consistency with Kotlin.

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
+A [`CredentialAttributes`](%gh-ic%/platform/credential-store/src/credentialStore/CredentialAttributes.kt) instance wraps credential metadata.
+To store a username, account name or other identity, set it along with the service name.
+
+<tabs group="languages">
+<tab title="Kotlin" group-key="kotlin">
+
+```kotlin
+import com.intellij.credentialStore.generateServiceName
+
+private const val SERVICE_NAME = "Password Storage Showcase"
```
karollewandowski: I would avoid using "Showcase" etc. in code, but try to use values closer to real world cases. IMHO, "My System" was OK.

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
-Examples:
+## Password Management
 
-| Subsystem       | Key              | Generated Service Name                               |
-|-----------------|------------------|------------------------------------------------------|
-| MyService API   | joe.doe          | IntelliJ Platform MyService API — joe.doe            |
-| Acme Repository | example.com/repo | IntelliJ Platform Acme Repository — example.com/repo |
+### Storing Passwords
 
+To store a `String`-based password, retrieve the `PasswordSafe` instance.
```
karollewandowski: I would improve this sentence. It sounds like the goal is to retrieve the PasswordSafe instance, but this is just the API that should be used.
Maybe:

```
To store a `String`-based password, use the `PasswordSafe` instance:
```
_notice the colon at the end_

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
-val password = credentials?.getPasswordAsString()
+> The `setPassword()` method of `PasswordSafe` is blocking and must not be invoked on EDT.
+> In Kotlin, it is recommended to declare a [Light Service](plugin_services.md#light-services) with a suspending function that calls `PasswordSafe` methods on the I/O dispatcher.
+{style="warning" title="Do not use `PasswordSafe` on EDT"}
+
+The password is persisted in [OS-specific storage](#storage).
+
+### Retrieving Passwords
+
+To retrieve a stored password,
```
karollewandowski:
```suggestion
To retrieve a stored password:
```

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
+  }
+
+  suspend fun load(): String? = withContext(Dispatchers.IO) {
+    PasswordSafe.instance.getPassword(credentialAttributes)
+  }
+
+  companion object {
+    private val serviceName = generateServiceName("Credentials Showcase", "API Key")
+
+    private val credentialAttributes = CredentialAttributes(serviceName)
```
karollewandowski: I'm wondering whether it is recommended to store attributes in memory for the entire service lifecycle, as it is kind of sensitive data (it doesn't store any password or token, but still security names). The `CredentialAttributes()` constructor only stores fields and is cheap to call when needed. I see no point in storing it in a service.

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
+```kotlin
+suspend fun getAsync(attributes: CredentialAttributes): Ephemeral<Credentials>
+```
+
+Besides being coroutine-friendly, it returns "ephemeral" credentials that are valid only while the client is connected to the backend in the [Remote Development](https://www.jetbrains.com/help/idea/remote-development-overview.html) context.
+When the client disconnects, the credentials are erased, preventing further actions on the user's behalf.
 
-> `PasswordSafe.set()` is blocking and shouldn't be called on EDT.
+> The `getAsync()` and `Ephemeral` APIs are under development
+> and experimental.
```
karollewandowski:
```suggestion
> The `getAsync()` and `Ephemeral` APIs are under development and experimental.
```

topics/basics/persisting_sensitive_data.md
```diff
@@ -139,10 +397,10 @@ The default storage format depends on the OS.
 [linux]: https://specifications.freedesktop.org/secret-service-spec/latest/
 [linux2]: https://wiki.gnome.org/Projects/Libsecret
 
-Users can override the default behavior in <ui-path>Settings | Appearance & Behavior | System Settings | Passwords</ui-path>.
+The default behavior can be overridden in <ui-path>Settings | Appearance & Behavior | System Settings | Passwords</ui-path>.
```
karollewandowski: I would keep the reference to "users" here as it is immediately clear that it is about plugin runtime behavior.

topics/basics/persisting_sensitive_data.md
```diff
@@ -2,127 +2,385 @@
...
 </tab>
 </tabs>
 
-To remove stored credentials, pass `null` for the `credentials` parameter.
+## Retrieving Credentials in Remote Development Context
+<primary-label ref="2025.3"/>
+
+For [Remote Development](https://www.jetbrains.com/help/idea/remote-development-overview.html), `PasswordSafe` provides an alternative way to retrieve credentials.
+
+```kotlin
```
karollewandowski, line 375: Is there any way to call it from Java? If not, I would mention that achieving this is possible only with Kotlin.

karollewandowski (changes requested):

topics/basics/persisting_sensitive_data.md
novotnyr: Inlined and added import to align with Kotlin

topics/basics/persisting_sensitive_data.md
novotnyr: Added import.

topics/basics/persisting_sensitive_data.md
novotnyr: Changed to "My Password Storage" to align with the actual value in the OS-specific storage.

topics/basics/persisting_sensitive_data.md
novotnyr: Good point.

topics/basics/persisting_sensitive_data.md
novotnyr: Calling from Java is out of scope, as this is a `suspend fun`. I will leave the Java migration as an exercise for the reader.

novotnyr: Put a "Kotlin" mention both here and in the experimental red box callout.

topics/basics/persisting_sensitive_data.md
novotnyr: Good point, reverted.

topics/basics/persisting_sensitive_data.md
novotnyr: That was purely for brevity. But I am applying your suggestion.

topics/basics/persisting_sensitive_data.md
karollewandowski: I renamed it back to `My Service`. Passwords are stored for services (AWS Something, LinkedIn Blah, etc.). Service name `My Password Storage` is possible, but rare, and reader can think that they should name their services like `AWS Password Storage` rather than `AWS Something`.

topics/basics/persisting_sensitive_data.md
karollewandowski: Thanks. I also inlined a couple of `val credentialAttributes = CredentialAttributes(username)` variables for shorter snippets. If it introduced a reasonable variable with a name explaining something, I would keep it, but here these are just repetitions.

topics/basics/persisting_sensitive_data.md
karollewandowski: > I will leave the Java migration as an exercise for the reader.

We should not treat docs gaps as opportunities to exercise. People want quick solutions and information and we should give it to them if possible. I will add a TODO marker to not block this PR.

karollewandowski (approved):

PR #2133, opened 2026-07-31
code_formatting.md: Fix typo

rapropos:
I believe there is an "or" that wants to be a "for".


novotnyr:
Yes, indeed. Thank you for spotting this!

PR #2156, opened 2026-08-17
Fix remote topic event sender method name

picimako:
Both `ApplicationRemoteTopic` and `ProjectRemoteTopic` use a function named `broadcast()` instead of `send()`.


novotnyr:
Thank you for spotting this! It's indeed `broadcast()`.

PR #2184, opened 2026-09-12
Document the non-embedded JNA library module in the 2026.3 API changes

develar:
Since intellij-community commit 1e44486b48d16 ("refactor platform: load JNA outside the core classloader"), the `intellij.libraries.jna` module is a plain content module of the core plugin and no longer loads in the core classloader. A plugin that uses JNA must now declare `<module name="intellij.libraries.jna"/>` in `plugin.xml` and `bundledModule("intellij.libraries.jna")` in the Gradle build script.

The note also records the `JnaLoader.load(Logger)` signature change from commit bf1917cb8f75c, which moved the class into the same module.

The note also recommends the Foreign Function and Memory API over JNA: plugins for 2026.2 and later target Java 25, and the IDE already runs with `--enable-native-access=ALL-UNNAMED`.


novotnyr (approved):

karollewandowski:
@develar, did you consider deprecating and scheduling `JnaLoader.load(Logger)` for removal via the API status and adding a new `JnaLoader.load()` without the parameter?

There are external plugins using it, and removal of the parameter will require them to release two separate versions (pre-2026.3 and 2026.3+).

jreznot:
https://youtrack.jetbrains.com/issue/IJPL-256666/Breaking-API-change-JnaLoader.load

